PR #98, opened 2020-10-21
First commit of gem-xcc coupling to score/develop

phyboyzhang:


src/importpart3mesh.cc
```diff
@@ -308,5 +310,337 @@ void Part3Mesh3D::DistributePoints(const double* exterarr, const LO gstart,LO li
...
+    if(test_case==TestCase::t0){
+      std::cout<<"rank="<<p1->mype<<" "<<li-gstart<<'\n';
+      std::cout<<"mylk k="<<mylk0[li-gstart]<<" "<<mylk1[li-gstart]
+      <<" "<<mylk2[li-gstart]<<" "<<'\n';
+    }
+}
+
 
+//fixme: the equivalent routine in GEM looks not right
+inline LO Part3Mesh3D::search_zeta(const double dlength,const double length,const LO nlength,double tmp)
```
Damilare06, line 552: Doesn't the use of a non-void return type in this routine negate the `inline` specification used? not detrimental but the inline functionality isn't valid for non-void routines, so I learned

phyboyzhang:
@Damilare06  Please don't care about grammar. If it's not right, I'll fix it in the debugging process. If you don't have other issue, just approve it.

Damilare06 (approved):

PR #107, opened 2022-04-04
Omega_h testing for send/recv

cwsmith:
This PR adds the following tests that use rendezvous to send and receive omegah mesh data.  Note, both tests are hard coded to only work with the meshes listed in the source code/ctest commands.  Generalizing their functionality will be part of future PR.

- `test_ohClassPtnSendRecv` has one application process (serial mesh) that sends data to two rendezvous processes where the mesh is partitioned by classification.  The rendezvous processes receive the data, confirm that it is correct, and attach it to the mesh.
- `test_ohClassPtn_appRibPtn` differs from the `test_ohClassPtnSendRecv` in two ways; (1) the application has two processes where the mesh is partitioned by the recursive inertial bisection method, and (2) a 'reverse' send is performed from rendezvous to the application after the 'forward' send from the application to rendezvous.


test/test_ohClassPtn_appRibPtn.cpp
```diff
@@ -0,0 +1,201 @@
...
+  redev::AdiosComm<redev::GO> commR2A(MPI_COMM_WORLD, appRanks, rdv.getFromEngine(), rdv.getFromIO(), name+"_R2A");
+
+  redev::LOs appOutPermute;
+  ts::OutMsg appOut;
+  ts::InMsg appIn;
+
+  redev::GOs rdvInPermute;
+  CSR rdvOutPermute;
+  ts::OutMsg rdvOut;
+  ts::InMsg rdvIn;
```
jacobmerson, line 135: Generally (unless there is profiling suggesting otherwise), these variables should be as local as possible. Compiler can bring them outside the loop if necessary and make code more readable.

test/test_ohClassPtn_appRibPtn.cpp
```diff
@@ -0,0 +1,201 @@
...
+        msgs[appOutPermute[i]] = gids_h[i];
+      }
+      auto start = std::chrono::steady_clock::now();
+      commA2R.Pack(appOut.dest, appOut.offset, msgs.data());
+      commA2R.Send();
+      ts::getAndPrintTime(start,name + " appWrite",rank);
+    } else {
+      auto start = std::chrono::steady_clock::now();
+      const bool knownSizes = (iter == 0) ? false : true;
+      ts::unpack(commA2R,knownSizes,rdvIn);
```
jacobmerson, line 159: Here if your api returns rdvIn it'll be a bit nicer. In C++17 you get either RVO or move so returning a struct is generally not expensive.

test/test_ohClassPtn_appRibPtn.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+    REDEV_ALWAYS_ASSERT(j!=gids_h.size()); //found
+    const auto subIdx = count[iGids[j]]++;
+    const auto startIdx = perm.off[iGids[j]];
+    const auto offIdx = startIdx + subIdx;
+    perm.val[offIdx] = iInGids[i];
+  }
+}
+
+void prepareRdvOutMessage(Omega_h::Mesh& mesh, ts::InMsg const& in, ts::OutMsg& out, CSR& permute){
```
jacobmerson, line 55: return struct?

test/test_support.cpp
```diff
@@ -0,0 +1,146 @@
...
+  Omega_h::Write inVtxData(inVtxData_h);
+  mesh.add_tag(0,name,1,Omega_h::read(inVtxData));
+  mesh.sync_tag(0,name);
+}
+
+void unpack(redev::AdiosComm<redev::GO>& comm, bool knownSizes, InMsg& in) {
+  redev::GO* msgs;
+  comm.Unpack(in.srcRanks, in.offset, msgs, in.start, in.count, knownSizes);
+  in.msgs = redev::GOs(msgs, msgs+in.count);
+  delete [] msgs;
```
jacobmerson, line 82: I think this points out a flaw in the Unpack api in redev...that api should either return a vector of GO or at least a unique ptr which gives clear intent of ownership.

test/test_support.h
```diff
@@ -0,0 +1,99 @@
...
+void checkAndAttachIds(Omega_h::Mesh& mesh, std::string name, redev::GOs& vtxData, redev::GOs& rdvPermute);
+
+/**
+ * Return the index permutation of the input array (v) such that the array is
+ * sorted in ascending order.
+ * from https://stackoverflow.com/a/12399290
+ */
+template <typename T>
+std::vector<size_t> sort_indexes(const T &v) {
+  // initialize original index locations
```
jacobmerson, line 86: sortIndexes?

jacobmerson (commented):
Did a cursory look at things nothing in depth. One minor thing I'm wondering about is if it would be more clear to write partition? instead of Ptn for the various variable names.

In general I think some of the main functions might be a bit more clear if each step gets a named function. And the functions might be more clear if you return data instead of return parameters whenever possible.

Fixing this sort of stuff in the test code isn't critical although they will get used as templates for other people.

Probably the biggest issue I saw was the rdv unpack function returning unmanaged memory. In "modern" c++ the general rule is: 1) use smart pointer to convey ownership transfer/sharing 2) use raw pointer for non-owning but can be null data and use reference for non-owning data that cannot be null

test/test_ohClassPtn.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <Omega_h_mesh.hpp>
+#include <Omega_h_for.hpp>
+#include <redev_comm.h>
+#include "wdmcpl.h"
+#include "test_support.h"
+
+namespace ts = test_support;
+
+void prepareMsg(Omega_h::Mesh& mesh, redev::ClassPtn& ptn,
+    ts::OutMsg& out, redev::LOs& permute) {
```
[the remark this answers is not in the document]
cwsmith: If you have an Omega_h PR the improves the interface we can merge it into the SCOREC fork and submit it to the SNL repo to see if Dan is interested.

I'll add the permutation to the 'OutMsg' struct as it is closely related the its contents and work on returning the struct.  

Jumping ahead to your other comment, I wasn't aware of RVO.  Thank you.  At first glance, this blog post on RVO looks like it provides a simple demonstration of the use case of interest: https://shaharmike.com/cpp/rvo/ .  I'll refactor the code to return the structs as you suggested.

test/test_support.cpp
```diff
@@ -0,0 +1,146 @@
...
+  MPI_Allreduce(&time, &min, 1, MPI_DOUBLE, MPI_MIN, comm);
+  MPI_Allreduce(&time, &max, 1, MPI_DOUBLE, MPI_MAX, comm);
+  MPI_Allreduce(&time, &tot, 1, MPI_DOUBLE, MPI_SUM, comm);
+  avg = tot / nproc;
+}
+
+void printTime(std::string mode, double min, double max, double avg) {
+  std::cout << mode << " elapsed time min, max, avg (s): "
+            << min << " " << max << " " << avg << "\n";
+}
```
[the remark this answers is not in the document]
cwsmith: Good point.  In the test usage though, only rank zero calls this function so I think we are OK.

https://github.com/SCOREC/wdmapp_coupling/blob/a573ddcfdeded4767d9761f199ff4fbf76c374e7/test/test_support.h#L37

test/test_support.cpp
```diff
@@ -0,0 +1,146 @@
+#include "test_support.h"
+#include <Omega_h_file.hpp> //vtk::write_parallel
+#include <algorithm> // std::sort, std::stable_sort
+#include <mpi.h>
+
+namespace test_support {
+
+void writeVtk(Omega_h::Mesh& mesh, std::string name, int step) {
```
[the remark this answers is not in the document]
cwsmith: Thank you.  Changed to `string_view` in [9ee8cd0](https://github.com/SCOREC/wdmapp_coupling/pull/107/commits/9ee8cd09a0b7e17f254ea4139df99a3d329ae361).

test/test_support.h
cwsmith: Good catch.  Changed in [a4d77bf](https://github.com/SCOREC/wdmapp_coupling/pull/107/commits/a4d77bfba19d94bb954a67c69c8c7f9e2fa99792).

test/test_support.cpp
cwsmith: In [redev v1.3.0](https://github.com/SCOREC/redev/releases/tag/v1.3.0) the `Pack`, `Send`,`Unpack` sequence of calls was changed to `SetOutMessageLayout` (called once), `Send(<messageData>)`, `<messageData> = Recv()`, and `<messageLayoutStruct> = GetInMessageLayout()`.  The simple `pingpong` test in redev shows a forward and reverse communication using the new APIs: https://github.com/SCOREC/wdmapp_coupling/pull/107/commits/62da1f7b373909fce3020da3383811e2f482a472

The Omega_h tests here are updated to use the new redev APIs with commits [62da1f7](https://github.com/SCOREC/wdmapp_coupling/pull/107/commits/62da1f7b373909fce3020da3383811e2f482a472) to [64d4d14](https://github.com/SCOREC/wdmapp_coupling/pull/107/commits/64d4d149ab5719f1ec42f8004af1db223268f558).

test/test_ohClassPtn_appRibPtn.cpp
cwsmith: see https://github.com/SCOREC/wdmapp_coupling/pull/107#discussion_r844003471

test/test_ohClassPtn_appRibPtn.cpp
cwsmith: The redev communication interface was revised.  See https://github.com/SCOREC/wdmapp_coupling/pull/107#discussion_r844003471

test/test_support.cpp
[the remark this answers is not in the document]
cwsmith: 👍 Changed in [615231d](https://github.com/SCOREC/wdmapp_coupling/pull/107/commits/615231d1cb49fc8469c03d73265a088643b9217a).

test/test_ohClassPtn_appRibPtn.cpp
```diff
@@ -0,0 +1,201 @@
...
+namespace ts = test_support;
+
+struct CSR {
+  redev::GOs off;
+  redev::GOs val;
+};
+
+//creates the outbound (rdv->non-rdv) permutation CSR given inGids and the rdv mesh instance
+//this only needs to be computed once for each topological dimension
+void getOutboundRdvPermutation(Omega_h::Mesh& mesh, const redev::GOs& inGids, CSR& perm) {
```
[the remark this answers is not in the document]
cwsmith: Fixed in [625335e](https://github.com/SCOREC/wdmapp_coupling/pull/107/commits/625335e4b3d25de565d30ef090fa466b47253acd).

test/test_ohClassPtn_appRibPtn.cpp
cwsmith: Fixed in https://github.com/SCOREC/wdmapp_coupling/commit/625335e4b3d25de565d30ef090fa466b47253acd.

cmake/runTwoMpiJobs.cmake
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.15.0...3.21.0)
+macro(mpi_test)
```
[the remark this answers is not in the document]
cwsmith: I'm not sure why I used a macro here.  I probably was looking at an example somewhere an mindlessly followed it.  Besides providing args as variables, the most interesting benefit of functions (according to [this SO post](https://stackoverflow.com/questions/24297999/function-vs-macro-in-cmake)) is the better scoping control.

Since this cmake file is called in [script mode](https://github.com/SCOREC/wdmapp_coupling/blob/a71b1ce1380dbd78b80963b915812103616018fc/test/CMakeLists.txt#L48-L54) I don't think there is a risk of scoping issues.  For consistency though, I've switched it to a function in [a71b1ce](https://github.com/SCOREC/wdmapp_coupling/pull/107/commits/a71b1ce1380dbd78b80963b915812103616018fc).

cwsmith:
@jacobmerson Would you please take another quick look at this when you get a chance?

test/test_support.cpp
```diff
@@ -0,0 +1,145 @@
...
+  mesh.add_tag(0,ohStr,1,Omega_h::read(inVtxData));
+  mesh.sync_tag(0,ohStr);
+}
+
+OutMsg prepareAppOutMessage(Omega_h::Mesh& mesh, const redev::ClassPtn& partition) {
+  //transfer vtx classification to host
+  auto classIds = mesh.get_array<Omega_h::ClassId>(0, "class_id");
+  auto classIds_h = Omega_h::HostRead(classIds);
+  //count number of vertices going to each destination process by calling getRank - degree array
+  std::map<int,int> destRankCounts;
```
jacobmerson, line 88: I don't think it matters too much because of the limited scope of this code. But I heard a tip that when you end up putting comments like this. e.g. step1 do something, step2 do something else. It ends up making things nicer if you extract that out into an independent function.

jacobmerson (approved):
I think this looks much nicer using the return structs.

PR #110, opened 2022-06-22
Uniform grid point search

jacobmerson:
Cleaned up a tangled mess of commits to get these commits that have the code for doing the uniform grid point search.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,7 +6,7 @@ set(WDMCPL_SOURCES  wdmcpl.cpp)
 
 add_library(wdmcpl ${WDMCPL_SOURCES})
 add_library(wdmcpl::wdmcpl ALIAS wdmcpl)
-target_compile_features(wdmcpl PUBLIC cxx_std_20)
+target_compile_features(wdmcpl PUBLIC cxx_std_17)
 target_link_libraries(wdmcpl PUBLIC Omega_h::omega_h redev::redev MPI::MPI_CXX)
 
 find_package(Kokkos REQUIRED)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -2,14 +2,39 @@ set(WDMCPL_HEADERS
   wdmcpl.h
   )
 
-set(WDMCPL_SOURCES
-  wdmcpl.cpp
-        wdmcpl/coordinate.h)
+set(WDMCPL_SOURCES  wdmcpl.cpp)
 
 add_library(wdmcpl ${WDMCPL_SOURCES})
 add_library(wdmcpl::wdmcpl ALIAS wdmcpl)
 target_compile_features(wdmcpl PUBLIC cxx_std_20)
-target_link_libraries(wdmcpl PUBLIC Omega_h::omega_h redev::redev MPI::MPI_C)
+target_link_libraries(wdmcpl PUBLIC Omega_h::omega_h redev::redev MPI::MPI_CXX)
+
+find_package(Kokkos REQUIRED)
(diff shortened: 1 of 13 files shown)


src/CMakeLists.txt
```diff
@@ -2,14 +2,19 @@ set(WDMCPL_HEADERS
...
 add_library(wdmcpl::wdmcpl ALIAS wdmcpl)
 target_compile_features(wdmcpl PUBLIC cxx_std_20)
-target_link_libraries(wdmcpl PUBLIC Omega_h::omega_h redev::redev MPI::MPI_C)
+target_link_libraries(wdmcpl PUBLIC Omega_h::omega_h redev::redev MPI::MPI_CXX)
+
+find_package(Kokkos REQUIRED)
+add_library(point_search wdmcpl/point_search.cpp wdmcpl/uniform_grid.h wdmcpl/bounding_box.h)
+add_library(wdmcpl::point_search ALIAS point_search)
+target_compile_features(point_search PUBLIC cxx_std_17)
+target_link_libraries(point_search PUBLIC Omega_h::omega_h wdmcpl Kokkos::kokkos MPI::MPI_CXX)
```
cwsmith, line 16: With the `wdmcpl` dependency listed can MPI and Omega_h be removed?

src/wdmcpl/point_search.cpp
```diff
@@ -0,0 +1,220 @@
+#include "point_search.h"
+#include <Omega_h_mesh.hpp>
+#include <bitset>
+
+namespace wdmcpl
+{
+// Omega_h::Vector<3> barycentric_from_global(
```
cwsmith, line 7: Can this commented code be removed?  It looks like it was moved to the bottom of the file.

src/wdmcpl/point_search.cpp
```diff
@@ -0,0 +1,220 @@
...
+    if (q < 0) {
+      return false;
+    }
+  }
+  return true;
+}
+// Liang, You-Dong, and B. A. Barsky. “A New Concept and Method for Line
+// Clipping.” ACM Transactions on Graphics 3, no. 1 (January 1984): 1–22.
+// https://doi.org/10.1145/357332.357333.
+bool line_intersects_bbox(const Omega_h::Vector<2>& p0,
```
cwsmith, line 64: Does this function, and any of the others below, need to be called from the GPU/device?

test/CMakeLists.txt
```diff
@@ -101,7 +101,10 @@ tri_mpi_test(TESTNAME test_twoClientOverlap_d3d_28p
 # unit tests
 find_package(Catch2 2 REQUIRED)
 message(STATUS "Found Catch2: ${Catch2_DIR} (found version ${Catch2_VERSION})")
-add_executable(unit_tests unit_test_main.cpp test_coordinate_transform.cpp test_coordinate.cpp)
-target_link_libraries(unit_tests PUBLIC Catch2::Catch2 wdmcpl::wdmcpl)
+#add_executable(unit_tests unit_test_main.cpp test_coordinate_transform.cpp test_coordinate.cpp test_field_abstraction.cpp test_point_search.cpp test_bounding_box.cpp test_uniform_grid.cpp)
```
cwsmith, line 104: remove?

cwsmith (changes requested):
LGTM.  Note, I only did a quick scan.  Please let me know if I should look at anything more critically.

src/CMakeLists.txt
jacobmerson: probably. I think this dependency is backwards though. I should probably make point_search directly part of `wdmcpl::wdmcpl`

jacobmerson:
Good call on the comments. Also, I realized I didn't test on the device. So I will fix that up as well. Some of the stdlib functions I used probably don't work on the device.

Issue #114, opened 2022-07-11
test_twoClientOverlap: partitioning the server to 16 ranks


cwsmith:
Running the `test_twoClientOverlap` server with 16 ranks and following cpn file results in the assert `REDEV_ALWAYS_ASSERT(entToRank[ent] == owners[i]);` being triggered in append() in test_support.cpp:

https://github.com/SCOREC/wdmapp_coupling/blob/d17a306105a9f70de25cadbb8bd6145422d23c33/test/test_support.cpp#L159-L160

The failing check indicates that the model vertex and edge ownership, derived from the cpn file, isn't self consistent.

https://github.com/SCOREC/wdmapp_coupling/blob/d17a306105a9f70de25cadbb8bd6145422d23c33/test/test_support.cpp#L189

Thanks @philip-davis for reporting this.

```

58
1 0
2 0
3 0
4 0
5 1
6 1
7 1
8 1
9 1
10 2
11 2
12 2
13 2
14 2
15 2
16 3
17 3
18 3
19 3
20 3
21 3
22 4
23 5
24 5
25 6
26 6
27 7
28 7
29 8
30 9
31 9
32 10
33 11
34 11
35 12
36 12
37 12
38 12
39 12
40 12
41 13
42 13
43 13
44 13
45 13
46 14
47 14
48 14
49 14
50 14
51 14
52 15
53 15
55 15
57 15
58 15
```

cwsmith:
A vtk written out before mesh element migration with the destination rank of each element attached as a tag seems looks OK.  

![destRanks0-8](https://user-images.githubusercontent.com/1024438/180866459-e5c1a8fd-77ec-4151-a5df-df9269e9cff9.png)
![destRanks9-15](https://user-images.githubusercontent.com/1024438/180866460-68f15b98-67a4-4f39-8686-14bd393181fa.png)


Given that, there is either a problem in the MPI send-recv of the element indices from rank 0

https://github.com/SCOREC/wdmapp_coupling/blob/9a5900a5bcee0031fa636606c21a91222e53fb3e/test/test_support.cpp#L220-L229

https://github.com/SCOREC/wdmapp_coupling/blob/9a5900a5bcee0031fa636606c21a91222e53fb3e/test/test_support.cpp#L246-L251

or, more likely, a bug in the(my) `Omega_h::migrate` function:

[...]

PR #117, opened 2022-08-17
resolve classification partition issue

cwsmith:
This PR addresses #114.

~~Once the debug code is removed I'll change this to an active PR (not a draft).~~


test/test_proxy_coupling.cpp
```diff
@@ -295,9 +346,9 @@ struct SerializeServer
 void coupler(MPI_Comm comm, Omega_h::Mesh& mesh, std::string_view cpn_file)
 {
 
-  wdmcpl::Coupler cpl("proxy_couple", wdmcpl::ProcessType::Server, comm,
-                      setupServerPartition(mesh, cpn_file));
-  auto is_overlap_h = markMeshOverlapRegion(mesh);
+  auto cp = setupServerPartition(mesh,cpn_file);
+  wdmcpl::Coupler cpl("proxy_couple", wdmcpl::ProcessType::Server, comm, cp);
+  auto is_overlap_h = markServerOverlapRegion(mesh,cp);
```
jacobmerson, line 351: Note that the Coupler has taken a copy of the partition `cp` and the `markOverlapRegion` will be working with a different copy of the partition object.

You can get the reference to the redev partition object with: `std::get<redev::ClassPtn>(rdv.GetPartition());`. For the coupler we need to add a getter for either redev or the partition.

[See test_setup_classPtn.cpp in RDV](https://github.com/SCOREC/redev/blob/9a4432b510bb5f01b3ba6010e741c4068977ddd9/test_setup_classPtn.cpp#L19)

jacobmerson (approved):
I approved since with the current usage the partition copy isn't a bug. But, it may not be working as intended so we can discuss further if need be.

test/test_proxy_coupling.cpp
cwsmith: Good catch.  I missed (or forgot) that there was a `std::move` in the Coupler ctor.

I just pushed https://github.com/SCOREC/wdmapp_coupling/pull/117/commits/db4ce9fcf7e6a0be790bb81339db4e3179b4f6d3.  Thoughts?

jacobmerson (approved):

PR #147, opened 2024-11-30
xSDK Changes

Angelyr:
- Doxygen page
- Smoke test
- optional print statement
- print version


Doxyfile
```diff
@@ -0,0 +1,2736 @@
...
+
+DOXYFILE_ENCODING      = UTF-8
+
+# The PROJECT_NAME tag is a single word (or a sequence of words surrounded by
+# double-quotes, unless you are using Doxywizard) that should identify the
+# project for which the documentation is generated. This name is used in the
+# title of most generated pages and in a few other places.
+# The default value is: My Project.
+
+PROJECT_NAME           = "Pcms"
```
cwsmith, line 45: Should this be "PCMS"?

src/pcms/print.cpp
```diff
@@ -0,0 +1,20 @@
...
+  FILE* getStderr() { return pp_stderr; }
+
+  void setStdout(FILE* out) {
+    assert(out != NULL);
+    pp_stdout = out;
+  }
+
+  void setStderr(FILE* err) {
+    assert(err != NULL);
+    pp_stderr = err;
```
cwsmith, line 18: Is the `pp_` prefix from PUMIPic?

test/CMakeLists.txt
cwsmith, line 1: The testing 'helper' macros were moved to `testing.cmake`.  Was the intention to also move all the tests (i.e. the calls to functions like `dual_mpi_test(...)` and their supporting code) to `testing.cmake`?  I don't see `testing.cmake` included/referenced in the CMakeLists.txt file but may have missed it.

cwsmith (changes requested):
Thank you.  A few comments are below.

test/CMakeLists.txt
Angelyr: I moved those functions so I could reuse them for the smoke tests. I included the file in the main CMakeLists.txt file and that is also where I created the smoke test. That way the smoke test is still created when testing is disabled.

cwsmith: OK.  Sounds good.  Thank you.

cwsmith (approved):

jacobmerson (approved):

jacobmerson:
Thanks!

PR #151, opened 2025-01-18
interpolator base class, tests [breaking]

Fuad-HH:
the test is still not passing. pushed it to save it at current state


Fuad-HH:
These are already merged in #178.

PR #152, opened 2025-01-18
Update support to Kokkos 4.5

Angelyr:
Updated to support Kokkos 4.5
Removed multiple definition in header
Fixed failing tests


test/test_coordinate.cpp
```diff
@@ -46,10 +46,10 @@ TEST_CASE( "Coordinate value type works", "[coordinate]" ) {
   namespace stdex = std::experimental;
   pcms::CoordinateMDArray<double,Cartesian> a(2);
 
-  std::experimental::mdarray<
+  Kokkos::mdarray<
     double,
-    std::experimental::extents<int, std::experimental::dynamic_extent, 3>,
-    std::experimental::layout_right, Omega_h::Write<double> > bla(2);
+    Kokkos::extents<int, Kokkos::dynamic_extent, 3>,
+    Kokkos::layout_right, Omega_h::Write<double> > bla(2);
```
cwsmith, line 52: Was mdarray use supposed to be removed?

cwsmith (changes requested):
LGTM.  I saw a use of mdarray and left a comment there.

test/test_coordinate.cpp
Angelyr: This is a test case that is commented out so I ignored it

cwsmith: @jacobmerson Should that test be removed?

cwsmith (approved):

PR #162, opened 2025-03-13
Partition Coupler Client

Angelyr:
added partition to construct client for ease of use in fusion-io


jacobmerson:
We previously had something like this then we removed it so that the client would get the partition from the server. The reason for this is that if the client provides a partition that is not the same as the server's partition, then things will break in unexpected ways.

As we discussed, there is some reasonable rational to have the client/server merge which will requite this. However, we also need logic to check that the user isn't providing incompatible things.

PR #166, opened 2025-03-18
add ridge regularization in the normal equation

abhiyanpaudel:


src/pcms/interpolator/mls_interpolation.hpp
```diff
@@ -19,6 +19,7 @@ Write<Real> mls_interpolation(const Reals source_values,
                               const Reals source_coordinates,
                               const Reals target_coordinates,
                               const SupportResults& support, const LO& dim,
-                              const LO& degree, RadialBasisFunction bf);
+                              const LO& degree, RadialBasisFunction bf,
+                              double lambda_factor = 0);
```
jacobmerson, line 23: Can you add a docstring to this function explaining all of the inputs? And, what the lambda factor is?

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -419,7 +432,8 @@ void mls_interpolation(RealConstDefaultScalarArrayView source_values,
                        RealConstDefaultScalarArrayView target_coordinates,
                        const SupportResults& support, const LO& dim,
                        const LO& degree, Func rbf_func,
-                       RealDefaultScalarArrayView approx_target_values)
+                       RealDefaultScalarArrayView approx_target_values,
+                       double lambda_factor)
 {
   PCMS_FUNCTION_TIMER;
```
jacobmerson, line 438: Are there any limits on the range of the lambda? I.e., does it need to be positive? If so, we should check those preconditions with an assert here.

test/test_linear_solver.cpp
```diff
@@ -159,10 +160,10 @@ TEST_CASE("solver test")
     Kokkos::View<double**, Kokkos::HostSpace> expected_solution(
```
jacobmerson, line 160: why are the expected results changing so much here? Shouldn't we get the same values when lambda=0?

jacobmerson (commented):
A few quick questions/changes. Then I can merge this change. Does it seem to help with the quality of the interpolation at all?

jacobmerson:
We decided to remove the normal equation based routines because they are horribly conditioned and do not provide good solutions for most of our test data. The SVD and QR based methods are much more effective.

PR #167, opened 2025-03-21
mls_interpolation with SVD solver 

abhiyanpaudel:
This pull request introduces an SVD-based solver to improve robustness in mls interpolation routines. The solver has been tested across a variety of scenarios and performs reliably in most typical cases.
- Works well for standard and moderately conditioned problems
- For extreme or ill-conditioned cases, adjusting the tolerance parameter in the solver may be necessary to ensure stability and accuracy

This offers a more stable alternative to existing LU solver that improves numerical stability in extreme cases.


src/pcms/interpolator/mls_interpolation.cpp
```diff
@@ -213,5 +244,52 @@ int calculate_scratch_shared_size(const SupportResults& support,
...
+  HostRead<Real> host_y(read(y_coordinates));
+
+  const auto min_x = Omega_h::get_min(read(x_coordinates));
+  const auto min_y = Omega_h::get_min(read(y_coordinates));
+
+  const auto max_x = Omega_h::get_max(read(x_coordinates));
+  const auto max_y = Omega_h::get_max(read(y_coordinates));
+
+  printf(" [min_x, max_x] : [%12.6f , %12.6f]\n", min_x, max_x);
+  printf(" [min_y, max_y] : [%12.6f , %12.6f]\n", min_y, max_y);
```
cwsmith, line 276: Any print statements should probably be within a conditional that checks the verbosity level.  IIRC, we want to make PCMS an xSDK package and that requires a 'silent' mode that doesn't write to std[out|err]

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -447,86 +643,85 @@ void mls_interpolation(RealConstDefaultScalarArrayView source_values,
...
-
-  team_policy tp(nvertices_target, Kokkos::AUTO);
+    calculate_scratch_shared_size(support, ntargets, basis_size, dim);
 
-  int scratch_size = tp.scratch_size_max(0);
+  team_policy tp(ntargets, Kokkos::AUTO);
 
+  int scratch_size = tp.scratch_size_max(1);
+  printf("Scratch Size = %d\n", scratch_size);
+  printf("Shared Size = %d\n", shared_size);
```
cwsmith, line 652: see comment above about print statements

src/pcms/interpolator/pcms_interpolator_array_ops.cpp
```diff
@@ -0,0 +1,20 @@
+#ifndef PCMS_INTERPOLATOR_ARRAY_OPS_HPP
+#define PCMS_INTERPOLATOR_ARRAY_OPS_HPP
```
cwsmith, line 2: Is this supposed to be a header or cpp file?

src/pcms/interpolator/pcms_interpolator_logger.hpp
```diff
@@ -0,0 +1,127 @@
+#ifndef PCMS_INTERPOLATOR_LOGGER_HPP
+#define PCMS_INTERPOLATOR_LOGGER_HPP
```
cwsmith, line 2: https://github.com/SCOREC/pcms/blob/develop/src/pcms/print.h defines logging through SPDLOG.  I assume this should use/conform to that?

src/pcms/interpolator/pcms_interpolator_view_utils.hpp
```diff
@@ -0,0 +1,179 @@
...
+{
+  int size = array.size();
+
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team, size), [=](int i) {
+    OMEGA_H_CHECK_PRINTF(
+      array(i) >= 0,
+      "[Error:] Square root of a negative number is invalid!\n"
+      "value is %12.6f\n",
+      array(i));
+    array(i) = sqrt(array(i));
```
cwsmith, line 62: Using `Kokkos::sqrt(...)` should be a more portable solution.

test/CMakeLists.txt
```diff
@@ -111,6 +111,10 @@ if(Catch2_FOUND)
 	      test_mls_basis.cpp
 	      test_rbf_interp.cpp
 	      test_linear_solver.cpp
+		  test_normalisation.cpp
+		  test_spr_meshfields.cpp
+		  test_svd_serial.cpp 
+		  # test_serial_qr.cpp
```
cwsmith, line 117: please remove the commented line
it looks like the indentation is different

test/test_serial_qr.cpp
cwsmith, line 1: This file was commented out in the build system.  Should it be removed?

test/test_svd_serial.cpp
```diff
@@ -0,0 +1,290 @@
...
+#include "KokkosBatched_SVD_Decl.hpp"
+#include "KokkosBatched_SVD_Serial_Impl.hpp"
+#include <KokkosBlas2_serial_gemv_impl.hpp>
+#include <KokkosBatched_Gemm_Decl.hpp>
+#include <vector>
+#include <iostream>
+
+using namespace std;
+using namespace Omega_h;
+using namespace pcms;
```
cwsmith, line 16: see comment above

cwsmith (changes requested):
Looks good.  I left a handful of comments below.

src/pcms/interpolator/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ set(PCMS_FIELD_TRANSFER_HEADERS
     linear_interpolant.hpp
     multidimarray.hpp
     mls_interpolation.hpp
+	pcms_interpolator_view_utils.hpp
+	pcms_interpolator_logger.hpp
```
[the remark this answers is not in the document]
abhiyanpaudel: I don’t understand why the indentation looks off here, because when I check the same code on my local machine, 
indentation appears consistent. What could be the reason?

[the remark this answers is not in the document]
cwsmith: Guess: there is a tab character on those lines and the editor is rendering it with a different number of spaces than github.  If there is a tab character, its a good practice to replace it with spaces to avoid this problem of 'how many spaces is a tab'.

src/pcms/interpolator/mls_interpolation.cpp
abhiyanpaudel: These were debugging print statements. I removed them. Is there any conditional that's been already used in pcms?

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -96,6 +119,9 @@ int calculate_scratch_shared_size(const SupportResults& support,
  *   @param[in] p A reference to the coordinate struct
  *   @param[in,out] basis_vector The polynomial basis vector
  *
+ *   @note the basis_vector size is more than basis_size
+ *   to accomodate with the output from qr
```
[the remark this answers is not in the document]
abhiyanpaudel: no, I removed it

src/pcms/interpolator/pcms_interpolator_aliases.hpp
```diff
@@ -3,7 +3,7 @@
 
 #include <Kokkos_Core.hpp>
 #include "../arrays.h"
-
+#include <Omega_h_array_ops.hpp>
```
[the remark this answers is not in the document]
abhiyanpaudel: no, I got rid of it

src/pcms/interpolator/pcms_interpolator_array_ops.cpp
abhiyanpaudel: The function in this file was redundant. I deleted this file.

src/pcms/interpolator/pcms_interpolator_logger.hpp
abhiyanpaudel: I wasn’t aware of SPDLOG. It was  written for use within Kokkos team-level execution, so I’m not sure whether SPDLOG is supported or compatible.

test/test_spr_meshfields.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <Omega_h_build.hpp>
+#include <Omega_h_file.hpp>
+#include <Omega_h_library.hpp>
+#include <Omega_h_array_ops.hpp>
+#include <catch2/matchers/catch_matchers_floating_point.hpp>
+#include <vector>
+#include <iostream>
+
+using namespace std;
+using namespace Omega_h;
```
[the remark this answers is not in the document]
abhiyanpaudel: I fixed it

test/test_spr_meshfields.cpp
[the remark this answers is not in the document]
abhiyanpaudel: yes, it will be a useful test case.

I removed them

The KokkosKernels SVD solver has an issue related to the default tolerance value, which is set to 0. This causes the internal while loop to get stuck.  I experimented with different tolerance values — setting it to `1e-6` allows the test to pass for some points, but not all. When set to `1e-8`, it gets stuck again.

test/test_svd_serial.cpp
abhiyanpaudel: removed

src/pcms/interpolator/mls_interpolation.cpp
cwsmith: AFAIK, these are the logging functions:
https://github.com/SCOREC/pcms/blob/develop/src/pcms/print.h

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -119,12 +116,10 @@ Reals min_max_normalization(Reals& coordinates, int dim);
  *   @param[in] p A reference to the coordinate struct
  *   @param[in,out] basis_vector The polynomial basis vector
  *
- *   @note the basis_vector size is more than basis_size
- *   to accomodate with the output from qr
  *
  */
 KOKKOS_INLINE_FUNCTION
-void eval_basis_vector(const IntDeviceMatView& slice_length, const Coord& p,
+void eval_basis_vector(const IntDeviceMatView& slice_length, const double* p,
```
cwsmith, line 122: I think passing a std::vector or std::array would be better than the raw pointer.

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -171,7 +164,7 @@ void eval_basis_vector(const IntDeviceMatView& slice_length, const Coord& p,
  **
  */
 KOKKOS_INLINE_FUNCTION
-void normalize_supports(member_type team, Coord& pivot,
+void normalize_supports(const member_type& team, double* target_point,
```
cwsmith, line 167: same point as above regarding the raw pointer

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -248,13 +230,18 @@ template <typename Func,
           std::enable_if_t<std::is_invocable_r_v<double, Func, double, double>,
                            bool> = true>
 KOKKOS_INLINE_FUNCTION void compute_phi_vector(
-  const Coord& target_point, const ScratchMatView& local_source_points, int j,
+  const double* target_point, const ScratchMatView& local_source_points, int j,
```
cwsmith, line 233: same point as above regarding the raw pointer

test/CMakeLists.txt
```diff
@@ -104,18 +104,16 @@ if(Catch2_FOUND)
...
+          test_point_search.cpp
+          test_mls_basis.cpp
+          test_rbf_interp.cpp
+          test_linear_solver.cpp
+          test_normalisation.cpp
 		  test_spr_meshfields.cpp
-		  test_svd_serial.cpp 
-		  # test_serial_qr.cpp
-              )
+		  test_svd_serial.cpp)
```
cwsmith, line 116: There may be another indentation issue here.  It appears misaligned on github.

src/pcms/interpolator/pcms_interpolator_logger.hpp
cwsmith: OK.  Using [kokkos::printf](https://kokkos.org/kokkos-core-wiki/API/core/utilities/printf.html) will make this gpu logging more portable.

src/pcms/interpolator/mls_interpolation.cpp
```diff
@@ -11,20 +11,20 @@ struct RBF_GAUSSIAN
...
                          r_sq);
 
     // 'a' is a spreading factor/decay factor
     // the value of 'a' is higher if the data is localized
     // the value of 'a' is smaller if the data is farther
-    int a = 20;
+    int a = 2;
 
     double r = sqrt(r_sq);
     double rho = sqrt(rho_sq);
```
cwsmith, line 30: kokkos::sqrt would be better here.  I think there are a few other places with `math.h` functions should use the kokkos namespace versions instead.

test/CMakeLists.txt
```diff
@@ -104,14 +104,16 @@ if(Catch2_FOUND)
...
+          test_field_transfer.cpp
+          test_uniform_grid.cpp
+          test_omega_h_copy.cpp
+          test_point_search.cpp
+          test_mls_basis.cpp
+          test_rbf_interp.cpp
+          test_linear_solver.cpp
+          test_normalisation.cpp
+		  test_spr_meshfields.cpp
+		  test_svd_serial.cpp)
```
cwsmith, line 116: The indentation looks off in github.

cwsmith (changes requested):
Looks good.  I saw a few small changes.

Any suggestions on where the thwaites mesh should be uploaded?  I don't see any meshes in the repo.

src/pcms/interpolator/mls_interpolation.cpp
```diff
@@ -11,20 +11,20 @@ struct RBF_GAUSSIAN
...
+    OMEGA_H_CHECK_PRINTF(r_sq >= 0,
                          "ERROR: square of  distance should always be positive "
                          "but the value is %.16f\n",
                          r_sq);
 
     // 'a' is a spreading factor/decay factor
     // the value of 'a' is higher if the data is localized
     // the value of 'a' is smaller if the data is farther
-    int a = 20;
+    int a = 5;
```
jacobmerson, line 27: This should be a member of the `RBF_GAUSSIAN` functor. That way you could provide a default and let the user change it.

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -13,13 +13,20 @@
...
 #include <KokkosBlas1_team_dot.hpp>
 #include <Kokkos_StdAlgorithms.hpp>
-
 #include "pcms_interpolator_aliases.hpp"
 #include "adj_search.hpp"
-#include "../assert.h"
-#include "../profile.h"
+#include "pcms/assert.h"
+#include "pcms/profile.h"
+#include "pcms_interpolator_view_utils.hpp"
```
jacobmerson, line 20: use the absolute path for these so include files are found same way in installed and local builds.

src/pcms/interpolator/mls_interpolation_impl.hpp
jacobmerson: Please use the `Rank2View` here. That's a `mdspan` with two indexes tagged for the memory space.

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -359,42 +415,76 @@ void solve_matrix(const ScratchMatView& square_matrix,
...
-  int size = vector.extent(0);
-  Kokkos::parallel_for(Kokkos::TeamThreadRange(team, size),
-                       [=](int j) { vector(j) = value; });
+  team.team_barrier();
+
+  eval_rhs_scaling(team, weight, rhs_values);
+
+  // initilize U (orthogonal matrices)  array
+  ScratchMatView U(team.team_scratch(1), row, row);
+  fill(-5.0, team, U);
```
jacobmerson, line 444: Do you need to fill these arrays before use? It seems like this will be doing unnecessary work?

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -359,42 +415,76 @@ void solve_matrix(const ScratchMatView& square_matrix,
...
+  ScratchMatView temp_matrix(team.team_scratch(1), column, row);
+  fill(0, team, temp_matrix);
+
+  // initialise V S^-1 U^T
+  ScratchMatView vsigmaInvUtMul(team.team_scratch(1), column, row);
+  fill(0, team, vsigmaInvUtMul);
+
+  if (team.team_rank() == 0) {
+    KokkosBatched::SerialSVD::invoke(KokkosBatched::SVD_USV_Tag(), matrix, U,
+                                     sigma, Vt, work, 1e-6);
```
jacobmerson, line 468: Tolerance probably needs to be in the function interface since we will likely need to mess around with that depending on the data set.

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -359,42 +415,76 @@ void solve_matrix(const ScratchMatView& square_matrix,
...
+    calculate_shrinkage_factor(team, lambda, sigma);
+
+    auto Ut = find_transpose(team, U);
+
+    scale_and_adjust(team, sigma, Ut, temp_matrix); // S^-1 U^T
+
+    KokkosBatched::SerialGemm<
+      KokkosBatched::Trans::Transpose, KokkosBatched::Trans::NoTranspose,
+      KokkosBatched::Algo::Gemm::Unblocked>::invoke(1.0, Vt, temp_matrix, 0.0,
+                                                    vsigmaInvUtMul);
```
jacobmerson, line 480: I think Serial operation should be on the team thread 0. Otherwise, you are doing this operation on multiple threads.

src/pcms/interpolator/mls_interpolation_impl.hpp
```diff
@@ -359,42 +415,76 @@ void solve_matrix(const ScratchMatView& square_matrix,
...
+
+    KokkosBlas::SerialGemv<
+      KokkosBlas::Trans::NoTranspose,
+      KokkosBlas::Algo::Gemv::Unblocked>::invoke(1.0, vsigmaInvUtMul,
+                                                 rhs_values, 0.0,
```
jacobmerson, line 485: same comment as above.

test/test_spr_meshfields.cpp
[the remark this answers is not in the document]
jacobmerson: Please add the test meshes to https://github.com/SCOREC/pcms_testcases . I just transferred this from my personal git to SCOREC.

jacobmerson (commented):
long term, I'd like for the internal functions to take `Rank1View` and `Rank2View` instead of Kokkos or Omega_h specific types (Read, View, etc). This will allow user to pass in data from application layer easier.

This change does not need to be part of this pull request.

src/pcms/interpolator/mls_interpolation.hpp
```diff
@@ -7,18 +7,82 @@ using namespace Omega_h;
```
jacobmerson, line 7: Don't know how we missed this on the first review, but this needs to be removed. @Fuad-HH already reported errors from this.

The problem with this is you are forcing all downstream code to bring all of Omega_h into it's namespace and that will create naming conflicts. For example with `parallel_for`

jacobmerson:
Looks like there are some build failures. Possibly missing header files?

test/test_spr_meshfields.cpp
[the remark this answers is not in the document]
jacobmerson: see issue #175

jacobmerson (approved):

PR #170, opened 2025-04-05
Fix #169

### LINKED ISSUE #169: Tests failing on ADA89 machine (uniform grid search, field interpolation, intersection map)
While testing the current `develop` branch on ADA89 machine, I encountered failing tests:

- field interpolation (identical fields)
- construct intersection map
- uniform grid search

@jacobmerson  could you please take a look?

matthew-mccall:
This pull addresses the issues found in #169. `UniformGrid` attempted to use STL algorithms, which were host-only, in a device context, resulting in the issues found. The STL algorithm's have been replaced with hand-rolled loops so that the device functions of `UniformGrid` now behave correctly. I've also moved the check for whether the point is inside the triangle to before the checks for nearest edge or vertex.

Changes to `src/pcms/point_search.cpp`:

* Simplified the loop and conditional logic in `GridPointSearch::operator()` by moving the barycentric inside check earlier and adding a break statement to exit the loop once a valid triangle is found. [[1]](diffhunk://#diff-b6ad98d142d76056630fef89bd0d6d47d7f0dfb5b7944b7937c9410f08d479d0L289-R300) [[2]](diffhunk://#diff-b6ad98d142d76056630fef89bd0d6d47d7f0dfb5b7944b7937c9410f08d479d0L341-L346)

Changes to `src/pcms/uniform_grid.h`:

* Replaced `std::transform` with a simple for loop to calculate `distance_within_grid` in `UniformGrid::ClosestCellID`.
* Replaced `indexes.fill(-1)` with a for loop to initialize indexes in `UniformGrid::ClosestCellID`.
* Replaced `std::reverse` with a manual reverse loop in `UniformGrid::ClosestCellID` and `UniformGrid::GetCellBBOX`. [[1]](diffhunk://#diff-d8de715ae86c08eaa20cb7871151b3f34c1040ec243778554360116410e37e9fL30-R69) [[2]](diffhunk://#diff-d8de715ae86c08eaa20cb7871151b3f34c1040ec243778554360116410e37e9fL82-R103)
* Replaced `std::accumulate` with a for loop to calculate the stride in `UniformGrid::GetDimensionedIndex`.

### DIFF
--- a/src/pcms/point_search.cpp
+++ b/src/pcms/point_search.cpp
@@ -286,12 +286,17 @@ Kokkos::View<GridPointSearch::Result*> GridPointSearch::operator()(Kokkos::View<
     // create array that's size of number of candidates x num coords to store
     // parametric inversion
     for (auto i = candidates_begin; i < candidates_end; ++i) {
-      int triangleID = candidate_map.entries(i);
-      auto elem_tri2verts =
-        Omega_h::gather_verts<3>(tris2verts, triangleID);
+      const int triangleID = candidate_map.entries(i);
+      const auto elem_tri2verts = Omega_h::gather_verts<3>(tris2verts, triangleID);
       // 2d mesh with 2d coords, but 3 triangles
-     auto vertex_coords = Omega_h::gather_vectors<3, 2>(coords, elem_tri2verts);
-     auto parametric_coords = barycentric_from_global(point, vertex_coords);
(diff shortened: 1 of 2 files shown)


jacobmerson (commented):
It's not clear to me why these algorithms were failing. But, at least for reverse you should factor out a function since you do that multiple times and it makes the code more clear.

For the transforms I'm ok with the raw loops.

jacobmerson (approved):

PR #171, opened 2025-04-30
C++20 Perlmutter Test

Angelyr:
Added test for c++20 on perlmutter


cwsmith (approved):
Looks good.  Thank you.

This replaces the build/test that doesn't use C++20?

Should we make C++20 the required default in the CMakeLists.txt?

Angelyr:
And yes this replaces the build that uses earlier version of c++.

cwsmith:
Sounds good.  Thank you.

PR #174, opened 2025-05-12
add the repo-ref to the cache key

### LINKED ISSUE #173: build caches for CI/CD don't include version numbers
The build cache key needs to include the version number for the software. We ran into an issue where we updated the kokkos versions, but the build was still working off the old cached version.

jacobmerson:
fixes #173


Angelyr:
Good Idea! I used to manually clear the cache

PR #179, opened 2025-05-20
Clean unused

jacobmerson:
this pull request cleans up some unused code and replaces ScalarArrayView with Rank1/Rank2 views.


src/pcms/arrays.h
```diff
@@ -13,105 +11,38 @@ namespace detail
...
+template <typename ElementType, typename MemorySpace>
+using Rank1View = View<1, ElementType, MemorySpace>;
 
 template <typename ElementType, typename MemorySpace>
-using ScalarArrayView = Kokkos::mdspan<
-  ElementType, Kokkos::dextents<LO, 1>,
-  Kokkos::layout_right,
-  detail::memory_space_accessor<std::remove_reference_t<ElementType>,
-                                MemorySpace>>;
+using Rank2View = View<1, ElementType, MemorySpace>;
```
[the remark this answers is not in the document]
cwsmith: seems like copilot is making a legit suggestion

cwsmith (changes requested):
Looks good.  There is one copilot suggestion below that looks correct.

src/pcms/adapter/omega_h/omega_h_field.h
```diff
@@ -242,37 +237,29 @@ auto get_nodal_data(const OmegaHField<T, CoordinateElementType>& field)
...
+  auto coords = get_ent_centroids(field.GetMesh(), mesh_entity_to_int(field.GetEntityType()));
+  if (field.HasMask()) {
+    // FIXME dimension should be made runtime parameter
+    return detail::filter_array<typename decltype(coords)::value_type,
+                                coordinate_dimension>(coords, field.GetMask(),
+                                                      field.Size());
   }
+  return coords;
   // should never be here. Quash warning
   return Omega_h::Reals{};
```
cwsmith, line 254: it looks like these lines can be removed now

src/pcms/arrays.h
```diff
@@ -13,105 +11,38 @@ namespace detail
...
-template <typename ElementType, typename CoordinateSystem, typename MemorySpace,
-          size_t N = 1>
-using CoordinateArrayView = Kokkos::mdspan<
-  CoordinateElement<ElementType, CoordinateSystem>,
-  Kokkos::extents<LO, Kokkos::dynamic_extent, N>,
-  Kokkos::layout_right,
-  detail::memory_space_accessor<ElementType, MemorySpace>>;
+template <int Rank, typename ElementType, typename MemorySpace>
+using View =
+  Kokkos::mdspan<ElementType, Kokkos::dextents<LO, Rank>, Kokkos::layout_right,
```
cwsmith, line 34: Should this be layout left (left most index is contiguous, aka column major) for rank2 views on GPUs?

src/pcms/field.h
```diff
@@ -46,16 +45,71 @@ class FieldAdapter
...
   virtual void Deserialize(
-    ScalarArrayView<T, MemorySpace> buffer,
-    ScalarArrayView<const pcms::LO, MemorySpace> permutation) const = 0;
+    Rank1View<T, MemorySpace> buffer,
+    Rank1View<const pcms::LO, MemorySpace> permutation) const = 0;
   virtual std::vector<GO> GetGids() const = 0;
   virtual ReversePartitionMap GetReversePartitionMap(
     const redev::Partition& partition) const = 0;
 };
 */
```
cwsmith, line 57: Is this commented code important?

cwsmith (changes requested):
Looks good.  A few comments/questions are below.

src/pcms/adapter/omega_h/omega_h_field.h
jacobmerson: indeed.

src/pcms/arrays.h
jacobmerson: That's a good question. I'm interested if you have a suggestion on best way to move forward. I think the following are the options:
1. Specialize the alias so that: `layout_left` is used in a GPU memory space and `layout_right` is used in a CPU memory space
2. Expose the layout in the alias
3. Have different named alias for left and right layouts

I think we need to support two interfaces, one that takes a GPU pointer and one that takes a CPU pointer in the fields since at runtime the user may want to use either depending on where the data is coming from.

src/pcms/field.h
jacobmerson: No, I think accidentally snuck in from some of the work we have been doing to redefine base classes for the fields. I will remove it.

jacobmerson: This is actually useful. It represents the model interface that is needed for the fields. In the 2.0 version of the fields we have proper inheritance as the type erasure ended up adding a lot of unneeded complexity.

PR #187, opened 2025-07-28
1d/2d cubic spline interpolator

Sichao25:
Implemented 1D and 2D spline interpolators with corresponding test cases. The interpolators use `mdspan` as input to give users full control over memory allocation strategies, allowing compatibility with various data layouts. 

While there is significant room for improvement in areas such as performance optimizations and implementation enhancements (e.g., more flexible constructors, integration with `UniformGrid` class, increased parallelism, C++ details, documentation ... ), this represents an initial version that's ready for preliminary feedback to identify any fundamental concerns.


jacobmerson:
@Sichao25 I meged the `Rank1View` stuff.

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+    * \param x input grid points in x direction
+    * \param y input grid points in y direction
+    * \param values values of given grid points
+    */
+  ExplicitBiCubicSplineInterpolator(Rank1View<T, MemorySpace> x,
+                                    Rank1View<T, MemorySpace> y,
+                                    Rank1View<T, MemorySpace> values);
+  /** \brief Set up the internal working space and coefficients
+    */
+  void setup(
```
jacobmerson, line 399: should this be called in the constructor instead of being a separate function the user needs to call?

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+    * ibcxmin=0 | ibcxmax=0  --  not-a-knot boundary condition.
+    * ibcxmin=1 | ibcxmax=1  --  first derivative match at x(0) or x(-1).
+    * ibcxmin=2 | ibcxmax=2  --  second derivative match at x(0) or x(-1).
+    * ibcxmin=3 | ibcxmax=3  --  first derivative equal to zero at x(0) or x(-1).
+    * ibcxmin=4 | ibcxmax=4  --  second derivative equal to zero at x(0) or x(-1).
+    * ibcxmin=5 | ibcxmax=5  --  first derivative matches the first divided difference.
+    * ibcxmin=6 | ibcxmax=6  --  second derivative matches the second divided difference.
+    * ibcxmin=7 | ibcxmax=7  --  third derivative matches the third divided difference.
+    */
+  ExplicitCubicSplineInterpolator(Rank1View<T, MemorySpace> x,
```
jacobmerson, line 99: For the boundary conditions, would it make sense to use an enum instead of an integer? Usually, I find that to be more clear and less prone to mistakes.

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+  inGrid(T &xget, T &zxget, Rank1View<T, MemorySpace> x, const LO &nx, LO &ier);
+
+  /** \brief Solve the cubic spline coefficients
+    */
+  static KOKKOS_FUNCTION void solve_spline(Rank1View<T, MemorySpace> x,
+                                           const LO &nx,
+                                           Rank2View<T, MemorySpace> fspl,
+                                           const LO &ibcxmin, const T &bcxmin,
+                                           const LO &ibcxmax, const T &bcxmax,
+                                           Rank1View<T, MemorySpace> wk);
```
jacobmerson, line 181: I suspect these are not private/protected because they are kokkos functions. @cwsmith do you have a suggestion for how to indicate that these are not user-facing functions? I'm not sure that I'm happy with most of the things I've tried in the past.

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+ *   h = (x(i+1)-x(i))
+ *   p = (x-x(i))/h
+ *   q = 1-p
+ *   s(x)= q*D(1,i) + p*D(1,i+1) + 
+ *       (h**2/6)*{[q**3-q]*D(2,i) + [p**3-p]*D(2,i+1)}
+ * where D(1,i) and D(2,i) are the coefficients of the compact cubic spline.
+ */ 
+template <typename T, typename MemorySpace>
+class CompactCubicSplineInterpolator
+    : public ExplicitCubicSplineInterpolator<T, MemorySpace> {
```
jacobmerson, line 230: can you explain the choice for inheritance here?

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+    * \param x input grid points
+    * \param values values of given grid points
+    * \param ibcxmin bounday condition flag at x(0)
+    * \param bcxmin bounday condition data at x(0)
+    * \param ibcxmax bounday condition flag at x(-1)
+    * \param bcxmax bounday condition data at x(-1)
+    *
+    * \details The boundary condition flags can take the same values as in ExplicitCubicSplineInterpolator.
+    */
+  CompactCubicSplineInterpolator(Rank1View<T, MemorySpace> x,
```
jacobmerson, line 247: there are a number of similar questions/comments I have to the other interpolator class. I'm not going to re-comment on everything.

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+    * \param bcmin boundary condition data
+    * \param ibcmin boundary condition flag
+    * \param iflg2 flag indicating if the boundary conditions are
+    * \param nx number of grid points in x direction
+    *
+    * \details A homogeneous boundary condition  set A has the property that if 
+    * s1(x,y) satisfies A, then (for any real number c) c*S1 also satisfies A, 
+    * and, if both s1(x,y) and s2(x,y) satisfy A then so also does s1+s2.
+    */
+  static void correction_detect(Rank1View<T, MemorySpace> bcmin, LO &ibcmin,
```
jacobmerson, line 440: Why is this needed? It's not 100% obvious to me (it could be quite useful).

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+/** \brief Solve for 1d spline coefficients.
+ * 
+ * \param k_bc1 boundary condition flag at x(0)
+ * \param k_bcn boundary condition flag at x(-1)
+ * \param n number of grid points
+ * \param x input grid points
+ * \param f output coefficients of the cubic spline
+ * \param wk working space for the spline solver
+ *
+ * \details Originally developed by Wayne Houlberg in Fortran.
```
jacobmerson, line 857: this is copied from pspline? If so, please put the full reference.

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+  ibc_check(ibcymin, "2d spline", "ymin", -1, 7);
+  if (ibcymin >= 0)
+    ibc_check(ibcymax, "2d spline", "ymax", 0, 7);
+
+  isUniformAscending<T, MemorySpace>(x, 1.0e-3);
+  isUniformAscending<T, MemorySpace>(y, 1.0e-3);
+  this->x_ = x;
+  this->nx_ = x.extent(0);
+  this->y_ = y;
+  this->ny_ = y.extent(0);
```
jacobmerson, line 1661: why are these being set in `sanity_check`? also, I think these `nx` and `ny` shouldn't be stored at all.

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+  T element;
+  Kokkos::View<T *, MemorySpace> data_view(span.data_handle() + index, 1);
+  auto element_host = Kokkos::create_mirror_view(data_view);
+  Kokkos::deep_copy(element_host, data_view);
+  element = element_host(0);
+  return element;
+}
+
+template <typename T, typename MemorySpace>
+void isUniformAscending(Rank1View<T, MemorySpace> x, const T &ztol) {
```
jacobmerson, line 31: Can we use `is_sorted` instead? https://kokkos.org/kokkos-core-wiki/API/algorithms/std-algorithms/all/StdIsSorted.html

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2963 @@
...
+          if (bcmin(ix) != 0.0) {
+            local_flag = 1;
+          }
+        },
+        Kokkos::Max<LO>(iflg2));
+  }
+}
+
+template <typename T, typename MemorySpace>
+void ExplicitCubicSplineInterpolator<T, MemorySpace>::sanity_check(
```
jacobmerson, line 744: a function with the name `sanity_check` should only do checking and have no side effects to the class invariants.

src/pcms/interpolator/spline_interpolator.hpp
Sichao25: Just following up on this issue. mdspan doesn’t provide iterators, right? If so, would it be better to stick with a hand-coded loop here?.

jacobmerson:
@Sichao25 We should be ready to make this into a regular pull request (currently it is a draft)

Sichao25:
Thanks for the notification. Just did that.

src/pcms/interpolator/spline_interpolator.hpp
```diff
@@ -0,0 +1,2845 @@
...
+ * within a small tolerance of the range. If it is within the tolerance, then
+ * clamp it to the nearest boundary value. If it is outside the tolerance, then
+ * set the error flag ier to 1.
+ */
+template <typename T, typename MemorySpace>
+KOKKOS_INLINE_FUNCTION void clamp_to_grid_range(T xget, T &zxget,
+                                                Rank1View<T, MemorySpace> x,
+                                                LO nx, LO &ier) {
+  if (xget < x[0] || xget > x[nx - 1]) {
+    T zxtol = 4.0e-7 * std::max(std::abs(x[0]), std::abs(x[nx - 1]));
```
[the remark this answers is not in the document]
jacobmerson: Generally it's better to provide the user the ability to mess with tolerances. Comparing floating point numbers is surprisingly frought, so there isn't really a universal tolerance that works for everything.

jacobmerson (approved):

Issue #192, opened 2025-08-24
Add clang-tidy to CI/CD


jacobmerson:
clang-tidy should be added to the CI/CD to improve the code quality and catch errors early.

PR #199, opened 2025-09-20
Create clang-tidy workflow

Sichao25:
Addressed issue #192 .

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -1,0 +1,177 @@
+name: clang-tidy-code-quality-check
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  test-build:
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+    - name: Update packages
+      run: sudo apt-get update
+
+    - name: Install mpi
+      run: |
+        sudo apt-get install -yq mpich bear
+
+    - uses: actions/checkout@v4
+
+    - name: build Catch2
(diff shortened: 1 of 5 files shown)


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,177 @@
...
+      run:
+        shell: bash
+
+    steps:
+    - name: Update packages
+      run: sudo apt-get update
+
+    - name: Install mpi
+      run: |
+        sudo apt-get install -yq mpich bear
```
jacobmerson, line 24: Instead of using `Bear` I think you can directly create the JSON compilation database by setting `CMAKE_EXPORT_COMPILE_COMMANDS=ON` in the cmake configruation.

https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html

jacobmerson (commented):
I have a couple higher level questions:
1. Will this require rebuilding all of the dependencies, or does `cache: true` mean that we will be reusing those libraries. If those are all the same as the cmake workflow should we factor out the dependencies into a shared `.yml` file that can be shared between them?
2. In the output of the workflow I see a lot of warnings listed, but no output from those. Are those suppressed warnings?

Sichao25:
About question 1,  I think we can reuse those libraries, which means using the same CMake workflow in a shared YAML file is a good idea.

About question 2, my understanding is that those errors originate from external sources (e.g. system headers), not our codebase. [Here](https://github.com/Sichao25/pcms/actions/runs/17873995436/job/50832352811#step:19:49) is the output with additional explanation. I've disabled those explanations in this PR to provide cleaner output.

PR #202, opened 2025-09-26
Downgrade cmake-format to 0.6.11 in CI

Sichao25:
Downgrade cmake-format to 0.6.11 because it is the latest version in spack-packages.


Sichao25:
The initial commit failed due to default configuration changes between two versions. Explicitly specifying the configuration file resolved this problem.

One more observation is that cmake-format didn't execute after clang-format encountered errors. Now they are separated into two independent jobs to prevent this blocking behavior.

PR #210, opened 2025-10-29
Switch from MPICH to OpenMPI in CI

Sichao25:
Switch to OpenMPI in CI due to known MPICH compatibility issues ([discussion](https://bugs.launchpad.net/ubuntu/+source/mpich/+bug/2072338)) with Ubuntu 24.04. Additionally, introduce MPIEXEC_PREFLAGS to support oversubscribing when launching multiple cores.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -1,3 +1,4 @@
+# Now it is separated to avoid affecting MPICH is not compatible with Ubuntu-latest and bear is not compatible with OPENMPI. This workflow should be merged after either compatibility issue is resolved.
 name: Clang-Tidy Check
 on:
   push:
@@ -10,7 +11,7 @@ on:
       - develop
 
 jobs:
-  test-build:
+  clang-tidy-check:
     runs-on: ubuntu-latest
 
     defaults:


cwsmith (approved):

Sichao25:
I'm not certain what the default timeout is, but we currently have it set to 30 minutes [here](https://github.com/SCOREC/pcms/blob/develop/.github/workflows/cmake-test.yml#L15). Is it okay to extend this?

.github/workflows/cmake-test.yml
```diff
@@ -165,12 +164,11 @@ jobs:
...
 
     - name: build pcms
-      if: matrix.clang_tidy == 'OFF'
       uses: ./.github/actions/install-repo
       with:
         repo-name: 'pcms'
-        repo-path: 'SCOREC/pcms'
-        repo-ref: ''
+        repo-path: 'Sichao25/pcms'
+        repo-ref: 'yus/ci_test'
```
jacobmerson: shouldn't the path for this still be `SCOREC/pcms`?

jacobmerson (changes requested):

.github/workflows/cmake-test.yml
Sichao25: yes, thanks for pointing that out. That's for my local testing.

Issue #221, opened 2025-11-18
meshFields NOT Found: Upstream Packages Failing to Build with PCMS


Fuad-HH:
I was trying to build another project that depends on PCMS. But the generation step failed with this error:

```bash
CMake Error at pcms/install/lib64/cmake/pcms/pcms_core-targets.cmake:61 (set_target_properties):
  The link interface of target "pcms::core" contains:

    meshfields::meshfields

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  pcms/install/lib64/cmake/pcms/pcms-config.cmake:36 (include)
  CMakeLists.txt:505 (find_package)
```

I found a workaround by adding `find_package(meshFields)` in my own project's CMakeLists.txt. But this is not ideal. As discussed, it can be due to not having meshFields added as a dependency in the CMake configuration.

jacobmerson:
resolved in #226

PR #226, opened 2025-11-25
Add meshFields as an exported target

Fuad-HH:
- [x] Export dependencies in the configuration
- [x] GitHub action item to test if pcms can be successfully used as a dependency library
- [x] Remove extra paths provided.
- [x] Follow PUMI's method


Fuad-HH:
I could also add a different repository instead of adding the dependent project in the `yml` file. Please suggest if you know a better approach.

jacobmerson (approved):

.github/workflows/cmake-test.yml
```diff
@@ -221,3 +221,49 @@ jobs:
...
+          -Dpcms_DIR=${{ runner.temp }}/build-pcms/install/lib/cmake/pcms \
+          -DCatch2_DIR=${{ runner.temp }}/build-Catch2/install/lib/cmake/Catch2 \
+          -DOmega_h_DIR=${{ runner.temp }}/build-omega_h/install/lib/cmake/Omega_h \
+          -DmeshFields_DIR=${{ runner.temp }}/build-meshFields/install/lib/cmake/meshfields \
+          -Dredev_DIR=${{ runner.temp }}/build-redev/install/lib/cmake/redev \
+          -Dflcl_DIR=${{ runner.temp }}/build-kokkos-fortran-interop/install/lib/cmake/flcl \
+          -DADIOS2_DIR=${{ runner.temp }}/build-ADIOS2/install/lib/cmake/adios2 \
+          -Dperfstubs_DIR=${{ runner.temp }}/build-perfstubs/install/lib/cmake \
+          -DKokkos_DIR=${{ runner.temp }}/build-kokkos/install/lib/cmake/Kokkos \
+          -DKokkosKernels_DIR=${{ runner.temp }}/build-kokkos-kernels/install/lib/cmake/KokkosKernels/
```
jacobmerson: Are all of these needed? I would think only the PCMS path is needed.

jacobmerson:
@Sichao25 @cwsmith The goal of this PR is to add testing for linking against PCMS to fix issues like #221. Do we have similar testing in place for our other libraries?

The approach here seems quite reasonable to me, although other options like [`try_compile`](https://cmake.org/cmake/help/latest/command/try_compile.html) may also provide a mechanism to do this.

.github/workflows/cmake-test.yml
Fuad-HH: Not all of them are needed. I will remove those. But I have noticed that perfstubs is needed.

Sichao25:
As far as I know, we don't have similar tests.

cwsmith:
PUMI has a an example CMakeLists.txt for users:
https://github.com/SCOREC/core/blob/83dfbb2175bc9434f20959fc11b72a785e046c3d/doc/CMakeLists.txt

and CI builds the example to check that our cmake config file is working:

https://github.com/SCOREC/core/blob/83dfbb2175bc9434f20959fc11b72a785e046c3d/.github/workflows/cmake.yml#L113-L142

I like this approach and it would be a nice addition to other projects.

.github/workflows/cmake-test.yml
jacobmerson: Looks like the issue is the same with `perfstubs`. It also needs to be added as a dependency to the configuration file.

Fuad-HH:
Updated everything. Now, projects depending on pcms should not ask for any pcms dependency locations. The test will fail until the related pull requests are merged and the cache cleared here if necessary. Here are the related pull requests: https://github.com/SCOREC/redev/pull/55, https://github.com/SCOREC/meshFields/pull/84.

.github/workflows/cmake-test.yml
Fuad-HH: Done for both perfstubs and meshFields.

Fuad-HH:
Redev already has perfstubs location (https://github.com/SCOREC/redev/pull/55 merged). Tests in CI/CD need redev cache updated and it should pass.

Fuad-HH:
@cwsmith, could you please reset the cache here? And is there anything else needed for this pull request?

cwsmith:
@Sichao25 Can you help out with the cache reset? Under the 'Actions' tab in the 'Caches' menu item on the left it lists 119 caches.  Do you delete them all to 'clear' the cache? I'm guessing many of them are old.

Sichao25:
I typically only delete those needed for current tests, since unused ones should be automatically cleaned up. I'll reset and rerun.

Fuad-HH:
It worked. Thank you, @Sichao25 and @cwsmith.

jacobmerson (approved):

PR #228, opened 2025-11-25
Added GlobalDataInterface

hgangwar:
This PR introduces `redev::AdiosGlobalComm` to provide a global communication interface between the coupler and the applications using the existing ADIOS channel.
**Usage:**
```cpp
auto GDI = app->Add_GDI<pcms::GO>("global_comm", comm);
````
This creates a global communicator that can send and receive data, for example:
```cpp
app->BeginSendPhase();
GDI->Send(mean.data(), "mean", mean.size());
app->EndSendPhase();

app->BeginReceivePhase();
mean = GDI->Receive("mean", mean.size());
app->EndReceivePhase();
```
- The `test_GDI` test case verifies global communication and data transfer.
- CMake fix to ensure the header file is installed in the expected include directory.


jacobmerson:
@hgangwar and I discussed offline. The plan for this is to replace the GDI name with `Data`.

So, we will have something on the order of as follows:
```c++
app->AddData("mydata", ...);
app->BeginSendPhase();
app->SendData("mydata");
app->EndSendPhase();

app->BeginReceivePhase();
app->ReceiveData("mydata");
app->EndReceivePhase();
```

jacobmerson:
This was subsumed by #341. Feel free to reopen if that's not correct.

PR #220, opened 2025-11-17
Self hosted ci

jacobmerson:

PR #229, opened 2025-11-26
Create self hosted CI

Sichao25:
Create self-hosted CI using the environments from the PR #220  .

Note: This should be tested after the GPU debug changes #227  are merged.


jacobmerson:
Do we know this PR is not trying to launch jobs on the self-hosted runner? It looks like you essentially copy pasted what I had in #220. The only thing that comes to mind as a possibility is the name in the yaml doesn't match the filename.

Otherwise, I wonder if there is some repo setting we need to change.

Sichao25:
I haven't figured out the reason yet. Is our ASICS environment configured for all branches or just selected branches?

jacobmerson:
@Sichao25 I think I fixed why it wasn't running. It turns out that the yml file needed to be inside the workflows folder. Please try pushing an empty commit to make sure the workflow still runs when you push and it's not related to some user permissions.

Sichao25:
Thanks a lot for spotting that! Didn't realize it was using the wrong directory.

jacobmerson:
Looks like it's working. I got the request for the deployment. I'm going to mark this ready for review.

@cwsmith do you want to review this PR? I think it's all set, but since there may be some security implications I rather get a third set of eyes.

Sichao25:
I might need to update the configuration with the latest branch before merging.

Sichao25:
Blocked by https://github.com/SCOREC/omega_h/pull/196. Let me know if you have suggestions for a better approach.

Sichao25:
Does it matter if I squash those debug commits?

jacobmerson:
Go ahead and squash all the debug commits.

jacobmerson:
I will merge this as soon as I get approval from @cwsmith.

.github/workflows/self-hosted.yml
```diff
@@ -0,0 +1,230 @@
+name: self-hosted.yml
+
+on:
+  pull_request:
+    branches:
+      - develop
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
cwsmith: I didn't see an option on the Actions tab to manually run this.  Did I miss it?  If it does exist, does it trigger the 'approval' emails as it does when someone commits to a PR?

.github/workflows/self-hosted.yml
```diff
@@ -0,0 +1,230 @@
...
+          echo "Test Results:" > ./pr/message
+          echo "- Kokkos CUDA: ${{ steps.build_pcms_kokkos_cuda.outcome }}" >> ./pr/message
+          echo "" >> ./pr/message
+          echo "[(details)](https://github.com/${{github.repository}}/actions/runs/${{ github.run_id }})" >> ./pr/message
+      - name: Upload Result
+        if: ${{ !cancelled() }} #upload unless the job was cancelled
+        uses: actions/upload-artifact@v4
+        with:
+          name: pr
+          path: pr/
```
cwsmith: IIUC, this isn't being used.  Is that correct?

Sichao25: That's correct. I will remove it.

.github/workflows/self-hosted.yml
Sichao25: I found a [post ](https://github.com/orgs/community/discussions/58022#discussioncomment-8716372)from 2024 saying that `workflow_dispatch` needs to be merged into or run on the master/main branch before the option becomes available. I'm not certain if this is still case. The second question will likely need testing to verify.

cwsmith: Ok.  My vote would be to remove it unless we think it is critical to have.

Sichao25: @jacobmerson Are there any critical use cases that depend on the workflow_dispatch trigger?

jacobmerson:
My guess is this failure
```
/space/d_zxg05646/asics-actions-runner-linux-x64-2.323.0/_work/pcms/pcms/pcmsCI_2025-12-16-12-50/build-pcms/test/test_interpolation_fortran_api: error while loading shared libraries: libmpicxx.so.12: cannot open shared object file: No such file or directory
```
means that we may need to specify that we are linking against MPI in the fortran api here:
https://github.com/SCOREC/pcms/blob/d50b976cacde5bffac9b246eceb010bb26588d60/src/pcms/fortranapi/CMakeLists.txt#L25

Or possibly, it needs to be set to be a `PUBLIC` dependency of the `capi`?

It's interesting that the the normal github action runners didn't catch this.

Sichao25:
Yes, MPI needs to be added to LD_LIBRARY_PATH for that test to pass.

jacobmerson:
Shouldn't the mpi module add mpi to the `LD_LIBRARY_PATH`?

Sichao25:
It might not be the case by default if I understand correctly. [Here ](https://spack.readthedocs.io/en/latest/module_file_support.html#customize-environment-modifications)is the related doc.

.github/workflows/self-hosted.yml
jacobmerson: Not that comes to mind.

Sichao25: Okay, then I will remove it.

jacobmerson (approved):

PR #234, opened 2025-12-10
Use commit hash as cache key instead of ref

Sichao25:
This allows the cache to distinguish between different versions of  dependencies, avoiding the use of stale cached builds. In CI workflow files, dependencies remain pinned with tag refs for readability unless there's a specific reason to use a particular commit.

Does this approach make sense to you? @cwsmith @jacobmerson


jacobmerson:
I haven't gone through the details of the code yet, but this would enable us to re-use cache for nightly tests of meshFields, or other items that are not getting lots of commits daily, but we want to keep track of in case errors are introduced. So conceptually, I think this is a good idea.

jacobmerson (approved):

PR #246, opened 2025-12-15
 Replace lambda with functor

Sichao25:
Minor fix to replace lambda functions with functor objects to prevent launching Kokkos kernels with lambda from private class methods.

### DIFF
--- a/src/pcms/interpolator/interpolation_base.cpp
+++ b/src/pcms/interpolator/interpolation_base.cpp
@@ -144,6 +144,85 @@ void adapt_radii(unsigned min_req_supports, unsigned max_allowed_supports,
   Kokkos::fence();
 }
 
+struct ScanSupportIdxFunctor
+{
+  Omega_h::Write<Omega_h::LO> support_ptr_l;
+  Omega_h::Write<Omega_h::LO> num_supports;
+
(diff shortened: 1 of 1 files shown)


src/pcms/interpolator/interpolation_base.cpp
```diff
@@ -154,15 +233,8 @@ void MLSPointCloudInterpolation::fill_support_structure(
...
-    "scan", n_targets_,
-    KOKKOS_LAMBDA(const int& i, unsigned& update, const bool final) {
-      update += num_supports[i];
-      if (final) {
-        support_ptr_l[i + 1] = update;
-      }
-    },
-    total_supports);
+  ScanSupportIdxFunctor scanFunctor(support_ptr_l, num_supports);
+  Kokkos::parallel_scan("scan", n_targets_, scanFunctor, total_supports);
```
jacobmerson, line 237: Just a note for these functors you can construct in-line and don't need a temporary variable.

`Comment view  Kokkos::parallel_scan("scan", n_targets_, ScanSupportIdxFunctor{support_ptr_l, num_supports}, total_supports);`

jacobmerson (approved):

Issue #112, opened 2022-07-01
Exceptions for Error handling


jacobmerson:
In some ways the code we write isn't too sensitive to this since error handling typically means just crash if you did something dumb.

Since the APIs are much cleaner if we use function returns not for error codes, I would suggest that we make use of exceptions.

@cwsmith do you have strong feelings one way or another on this?

cwsmith:
Yeah, I don't think I've been involved with any package that 'tried' to do something knowing it may not work and attempt to recover from that failure.  Along those lines, I don't have any significant experience with error handling (beyond `asserts` etc.) so I'm pretty open to any option that is relatively simple to work with.

Do you have a small example of the exception based failure code?  I've seen libraries that suggest placing try/catch blocks at the top level if the user wants to print meaningful error messages.  Is that the idea here?

The xSDK folks have a 'recommended' policy, R3, regarding error handling here that we can consider:

https://figshare.com/articles/journal_contribution/xSDK_Community_Package_Policies/4495136

I like this idea:

> For example, all routines may, by default, return an error code with the option of changing it to
generate an abort on the error (for running in the debugger).

Sichao25:
As a starting point, I've created a base exception class and error handling functions to control error behavior. My expectation is that we can use the functions across the codebase, with behavior controlled globally through configuration. Would like to hear any comments and suggestions before moving on.

PR #249, opened 2025-12-17
Add base exception class and error handling utilities

Sichao25:
Based on the discussion in issue #112 , introduces a base exception class for PCMS error handling and adds utility functions to control error behavior. This provides a foundation for error handling across the codebase.


jacobmerson:
I think this is much more complex than it needs to be. The compiler will insert a `std::terminate` if the code is compiled with `-fnoexcept`. My suggestion would be that we just create a basic pcms exception type that can live in the utility library I defined in #252.

An example of what this exception type should look like can be found here: https://github.com/SCOREC/mumfim/blob/develop/src/mumfim/exceptions.h

jacobmerson:
Now that Kokkos requires C++20, I feel comfortable using C++20 in PCMS. In light of this would it make sense to capture the location of the error with `https://en.cppreference.com/w/cpp/utility/source_location.html`?

src/pcms/assert.h
```diff
@@ -44,6 +47,31 @@
 
 namespace pcms
 {
+
+class exception : public std::exception
```
jacobmerson: is there a specific reason to use codes here? I think the typical thing to do is define a base Exception type  for PCMS based on the std::exception, then we can have additional exceptions that inherit from that. I.e., IOError, etc. Then for the CAPI, you typically have a function that converts from the exception type to an error code that gets returned from the functions.

jacobmerson (approved):

Sichao25:
@jacobmerson Can I merge this if there are no further comments?

jacobmerson (approved):

PR #252, opened 2025-12-21
move common functionality to a utility library

jacobmerson:
This reorganization will help decouple components. For example, the vmec field needs to use interpolation. With the utility libraries moved to a lower level, we can avoid a circular dependency between pcms::core pcms::interpolator. That is, pcms::core can utilize pcms::interpolator. This is really the first step in a broader re-organization of the pcms::sources to be more modular.


jacobmerson:
I think the clang-tidy failure here is actually an issue with redev. See: https://github.com/SCOREC/redev/issues/56

jacobmerson:
Note:
1. self hosted runner is currently failing with:
  ```cudaGetDeviceCount(&count) error( cudaErrorInitializationError): initialization error /space/d_zxg05646/asics-actions-runner-linux-x64-2.323.0/_work/pcms/pcms/pcmsCI_2025-12-21-18-05/kokkos/core/src/impl/Kokkos_Core.cpp:134```
  see: https://github.com/SCOREC/pcms/actions/runs/20416693526/job/58661403802?pr=252#step:4:7930
  I think this is the intermittent failure that @Sichao25 @cwsmith have been seeing on the self-hosted runner. We will try re-running later.
2. @jacobmerson (note to self) at this point, don't force push as other branches are based on this.

Sichao25 (commented):
Looks good to me!

jacobmerson:
@Sichao25 this is mostly reorganization of the files to start trying to clean up the source directory. Let me know if you have comments / issues.

PR #254, opened 2026-01-13
test cmake-based approach to run clang-tidy

jacobmerson:


jacobmerson:
Just experimenting to see if we can use this to get rid of the Bear step that is super slow. The initial test indicates that at least some files are not properly picking up the `mpi.h` header which is probably why we were using Bear in the first place...

jacobmerson:
This approach seems to be significantly faster. Roughly 4 min 30 sec to run clang-tidy vs. 5 min 45sec for Bear + 9 min to run clang-tidy with the current workflow.

jacobmerson:
@Sichao25 @cwsmith Finally got this working without Bear. The strategy here has significant time savings with respect to running clang tidy because it avoids the use of Bear. The previous full clang-tidy step (including build pcms, etc.) is something like 16 minutes and this new one is 6m 30s.

cwsmith (approved):
Nice!

Sichao25 (commented):
Looks good to me!

PR #258, opened 2026-02-06
Python API for Unstructured Mesh Field Transfer

Sichao25:
Introduce Python APIs for transferring fields between unstructured and structured meshes. It includes:
- Definition of uniform grid field and layout
- Python bindings for essential transfer functions


src/pcms/pythonapi/test_field_copy.py
```diff
@@ -0,0 +1,84 @@
+import py_pcms
```
jacobmerson: Can we make the python module name just `pcms` instead of `py_pcms`? This might require writing a `__init__.py` since the compiled shared library probably still needs to be called `py_pcms.so` to avoid naming conflicts.

src/pcms/pythonapi/test_field_copy.py
```diff
@@ -0,0 +1,84 @@
...
+        nx, ny, nz, 
+        False
+    )
+    print(f"  Built {dim}D mesh with {nx}x{ny}x{nz} elements")
+    print(f"  Mesh type: {type(mesh)}")
+    print(f"  Mesh object: {mesh}")
+
+    # Create layout
+    print("  About to create layout...")
+    layout = py_pcms.create_lagrange_layout(
```
jacobmerson: default to `num_components=1` and `Cartesian` for the coordinate system.

src/pcms/pythonapi/test_field_copy.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Create original field and set data
+    original = layout.create_field()
+    print("  Created original field")
+    original.set_dof_holder_data(ids)
+    print("  Set data in original field")
+
+    # Create copied field and copy data
+    copied = layout.create_field()
+    py_pcms.copy_field2_Real(original, copied)
```
jacobmerson: take of the `2` from the name. Also, we should handle the types internally. So in the interface we take in the field and in the interface layer we dispatch to the appropriate type.

src/pcms/pythonapi/bind_uniform_grid_field.cpp
```diff
@@ -0,0 +1,230 @@
...
+      auto contiguous_data = py::array_t<Real>(data);
+      auto data_view = numpy_to_view<Real>(contiguous_data);
+      // Create const view wrapper
+      Rank1View<const Real, HostMemorySpace> const_view(data_view.data_handle(), data_view.size());
+      self.SetDOFHolderData(const_view);
+    },
+    py::arg("data"),
+    "Set the DOF holder data")
+
+    .def("to_mdspan", [](const UniformGridField<2>& self) {
```
jacobmerson: On the python side we should change the name to `to_numpy` with everything else the same.

I also think there may be a way to avoid the copy: https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html

src/pcms/pythonapi/test_file_io.py
```diff
@@ -0,0 +1,452 @@
...
+    
+    # Get initial mesh properties
+    nverts_orig = mesh.nverts()
+    nelems_orig = mesh.nelems()
+    dim_orig = mesh.dim()
+    print(f"Original mesh: dim={dim_orig}, nverts={nverts_orig}, nelems={nelems_orig}")
+    
+    # Create temporary directory for test files
+    test_dir = "test_data"
+    os.makedirs(test_dir, exist_ok=True)
```
jacobmerson: Since you cleanup the directory as part of this test, you should throw an error if it already exists to avoid deleting something that we did not create.

src/pcms/pythonapi/test_file_io.py
```diff
@@ -0,0 +1,452 @@
...
```
jacobmerson: same comment on creation/cleanup of test directory.

src/pcms/pythonapi/test_file_io.py
```diff
@@ -0,0 +1,452 @@
...
```
jacobmerson: same cleanup issue.

jacobmerson (changes requested):
@Sichao25 one small comment so that we avoid potentially deleting files that we did not create. Otherwise this looks good to go.

Thanks!

src/pcms/pythonapi/test_file_io.py
Sichao25: Just updated to use the default behavior (If the target directory already exists an OSError is raised if exist_ok is False otherwise not.)

jacobmerson (approved):

PR #259, opened 2026-02-14
Handle out-of-bounds coordinates during evaluation

Sichao25:
Handle out-of-bounds cases by throwing an exception or filling a given value. NEAREST_BOUNDARY mode (map to nearest elements) is not yet implemented.

### DIFF
--- a/src/pcms/adapter/omega_h/omega_h_field2.cpp
+++ b/src/pcms/adapter/omega_h/omega_h_field2.cpp
@@ -166,15 +166,26 @@ struct OmegaHField2LocalizationHint
     std::vector<size_t> valid_point_indices;
     std::vector<size_t> missing_point_indices;
 
-    for (size_t i = 0; i < search_results.size(); ++i) {
-      auto [dim, elem_idx, coord] = search_results(i);
-      bool is_valid = (static_cast<int>(dim) == mesh.dim()) &&
-                      (elem_idx >= 0) && (elem_idx < mesh.nelems());
-
-      if (is_valid) {
+    if (mode_ == OutOfBoundsMode::ERROR) {
+      // Error mode - throw error immediately if any point is out of bounds
+      for (size_t i = 0; i < search_results.size(); ++i) {
+        auto [dim, elem_idx, coord] = search_results(i);
+        bool is_missing =
+          (static_cast<int>(dim) != mesh.dim()) || (elem_idx < 0);
+        PCMS_ALWAYS_ASSERT(!is_missing && "Points found outside mesh domain");
         valid_point_indices.push_back(i);
-      } else {
-        missing_point_indices.push_back(i);
+      }
+    } else {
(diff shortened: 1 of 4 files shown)


src/pcms/field.h
```diff
@@ -76,6 +76,13 @@ struct LocalizationHint
   std::shared_ptr<void> data = nullptr;
 };
 
+enum class OutOfBoundsMode
+{
+  ERROR, // Throw error when points are out of bounds
+  FILL,  // Fill out-of-bounds points with a fill value
+  CLAMP  // Clamp to nearest boundary cell (extrapolate)
```
jacobmerson: Can we call this NEAREST_BOUNDARY? `Clamp` means "saturate" to some high or low value. Such as: https://en.cppreference.com/w/cpp/algorithm/clamp.html

jacobmerson:
@hgangwar please review this when you get a chance. Changes here are what we discussed about handling points outside the domain.

src/pcms/adapter/omega_h/omega_h_field2.cpp
```diff
@@ -161,49 +158,96 @@ struct OmegaHField2LocalizationHint
...
+    : mode_(mode), fill_value_(fill_value), num_valid_(0), num_missing_(0)
   {
-    Kokkos::View<LO*, HostMemorySpace> elem_counts("", mesh.nelems());
+    // First pass: count valid and invalid points
+    std::vector<size_t> valid_point_indices;
+    std::vector<size_t> missing_point_indices;
 
     for (size_t i = 0; i < search_results.size(); ++i) {
       auto [dim, elem_idx, coord] = search_results(i);
+      bool is_valid = (static_cast<int>(dim) == mesh.dim()) &&
```
jacobmerson: Just check if the index is negative. A negative index means the point is outside.

src/pcms/adapter/omega_h/omega_h_field2.cpp
```diff
@@ -161,49 +158,96 @@ struct OmegaHField2LocalizationHint
...
+      } else {
+        missing_point_indices.push_back(i);
+      }
+    }
+
+    num_valid_ = valid_point_indices.size();
+    num_missing_ = missing_point_indices.size();
+
+    // Handle missing points based on mode
+    if (num_missing_ > 0 && mode_ == OutOfBoundsMode::ERROR) {
```
jacobmerson: rather than waiting to process points, let's throw an error as soon as you get an invalid point.

jacobmerson (changes requested):

hgangwar:
@jacobmerson , the test case validates exactly the functionality we need. As I understand it, any out of bounds vertex's element index is marked as something `< 0`, and handled accordingly via `OutOfBoundsMode`. I will verify it further with the convergence testcase.

jacobmerson (approved):

PR #260, opened 2026-02-19
Mirror source tree for adapter headers

hgangwar:
Install adapter headers via `DIRECTORY` install.

## Changes

* Defined `ADAPTER_HEADERS`.
* Installed headers from `ADAPTER_HEADERS` into their respective directories.
* Included `create_field.h` in examples for external-usage-example usage testing.

### DIFF
--- a/examples/external-usage-example/main.cpp
+++ b/examples/external-usage-example/main.cpp
@@ -1,5 +1,6 @@
 #include <iostream>
 #include <pcms/utility/print.h>
+#include <pcms/create_field.h>
 
 int main()
 {
--- a/examples/external-usage-example/main.cpp
+++ b/examples/external-usage-example/main.cpp
@@ -1,7 +1,7 @@
 #include <iostream>
 #include <pcms/utility/print.h>
 #include <pcms/create_field.h>
-
+#include <pcms/capi/interpolator.h>
 int main()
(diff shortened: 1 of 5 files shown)


hgangwar:
@jacobmerson

jacobmerson:
We had a breif discussion offline. We are going to investigate converting our header definitions over to `FILE_SETS` which should maintain the structure.

https://cmake.org/cmake/help/latest/command/install.html#id5

This approach looks like it may also enable us to set up targets that check that the header files can be correctly included, which would be a nice test to include to our CI/CD testing.

https://cmake.org/cmake/help/latest/prop_tgt/VERIFY_INTERFACE_HEADER_SETS.html#prop_tgt:VERIFY_INTERFACE_HEADER_SETS

jacobmerson (approved):
@hgangwar this looks good. Are there other header lists that should be converted to use this method? If so, lets implement that and then make this PR not a draft.

hgangwar:
Added `field_layout_communicator.h` to `PCMS_HEADERS`, required by `field_communicator2.h`.

jacobmerson:
I will merge when remaining headers are added to respective `FILE_SET` so we have a consistent handling of headers in our CMake.

PR #273, opened 2026-03-06
New field update

Sichao25:
Wrapping up the update in #250 
- Use template for omega_h_field2
- Update layout communicator
- Rename CreateField to CreateFieldReal


src/pcms/adapter/meshfields/mesh_fields_adapter2.h
```diff
@@ -5,29 +5,287 @@
...
+#include "pcms/utility/profile.h"
 #include "pcms/field.h"
 #include "pcms/coordinate_system.h"
 #include "pcms/point_search.h"
 
 namespace pcms
 {
-
+template <typename T>
 class MeshFieldBackend
```
jacobmerson, line 20: @cwsmith does MeshFields support all types or is it restricted like Omega_h tags?

src/pcms/adapter/omega_h/omega_h_field2.h
```diff
@@ -5,29 +5,287 @@
...
+    if (is_final) {
+      offsets_(i) = partial;
+    }
+    partial += elem_counts_(i);
+  }
+};
+
+struct CountPointsPerElementFunctor
+{
+  Kokkos::View<LO*> elem_counts_;
```
jacobmerson, line 123: The Adapter should throw an error if used with a 3D mesh until we resolve all these hardcoded 2D mesh items.

src/pcms/adapter/omega_h/omega_h_field2.h
```diff
@@ -5,29 +5,287 @@
...
+  // indices are the index of the original point (for valid points)
+  Kokkos::View<LO*, HostMemorySpace> indices_;
+  // indices of points not found in mesh
+  Kokkos::View<LO*, HostMemorySpace> missing_indices_;
+};
+
 // TODO template over possible OmegaHField2Types
-class OmegaHField2 : public FieldT<Real>
+template <typename T>
+class OmegaHField2 : public FieldT<T>
```
jacobmerson: I think we should actually call this the MeshFields adapter. That way we can later add an Omega_h adapter that can be built without linking against MeshFields that only supports vertex-based linear fields (which is trivial based on barycentric coordinate definition that we already compute).

src/pcms/adapter/meshfields/mesh_fields_adapter2.h
```diff
@@ -36,35 +294,308 @@ class OmegaHField2 : public FieldT<Real>
...
+      KOKKOS_LAMBDA(LO i) { values[hint.missing_indices_(i)] = fill_val; });
+  }
+}
+
+template <typename T>
+inline void OmegaHField2<T>::EvaluateGradient(
+  FieldDataView<T, HostMemorySpace> /* unused */)
+{
+  // TODO when moved to PCMS throw PCMS exception
+  throw std::runtime_error("Not implemented");
```
jacobmerson: update to PCMS exception type. Please open a feature request to add this so it is in the issue tracker. We should be able to easily support computing gradients from MeshField shape function derivatives.

src/pcms/adapter/omega_h/omega_h_field2.h
```diff
@@ -36,35 +294,308 @@ class OmegaHField2 : public FieldT<Real>
...
+
+template <typename T>
+std::array<Kokkos::View<LO*>, 4> OmegaHField2<T>::create_dim_masks(
+  Omega_h::Mesh& mesh, const Kokkos::View<Omega_h::I8*> elem_mask,
+  std::array<int, 4> nodes_per_dim)
+{
+  PCMS_ALWAYS_ASSERT(mesh.nelems() == elem_mask.extent(0));
+  std::array<Kokkos::View<LO*>, 4> masks;
+
+  struct Helper
```
jacobmerson: This seems like it's indicating a problem in one of our definitions if we need to do this.

src/pcms/adapter/omega_h/omega_h_field_layout.cpp
```diff
@@ -182,6 +193,21 @@ OmegaHFieldLayout::OmegaHFieldLayout(Omega_h::Mesh& mesh,
     offset += mesh.nents(i);
   }
 
+  if (mask.exists()) {
+    PCMS_ALWAYS_ASSERT(false);
+    // using ExecutionSpace = HostMemorySpace::execution_space;
```
jacobmerson: what is this commented code? At a minimum, we need to document why it's here/commented out. Better, we should either fix it or remove it.

This looks like it is in place to support the field masking.

Note: I just had a discussion with @hgangwar about the API for masking and long term, we will want to pass in geometric classification information to the field adapters rather than bitmasks that get constructed from geometric classification externally.

src/pcms/adapter/omega_h/omega_h_field_layout.cpp
```diff
@@ -200,9 +226,9 @@ OmegaHFieldLayout::OmegaHFieldLayout(Omega_h::Mesh& mesh,
   gids_host_ = Omega_h::HostWrite<Omega_h::GO>(gids_);
 }
 
-std::unique_ptr<FieldT<Real>> OmegaHFieldLayout::CreateField() const
+std::unique_ptr<FieldT<Real>> OmegaHFieldLayout::CreateFieldReal() const
```
jacobmerson: At the time when @joyal-mathew implemented this, the idea of creating a field from the layout made a lot of sense to me in terms of simplification of the downstream APIs. However, it's not clear to me that a layout always has sufficient information to construct a field.

Moving forward, we should investigate an alternative API that externalizes this capability.

Note: this can be addressed in a later PR.

src/pcms/adapter/omega_h/omega_h_field_layout.h
```diff
@@ -19,7 +19,12 @@ class OmegaHFieldLayout : public FieldLayout
                     int num_components, CoordinateSystem coordinate_system,
                     std::string global_id_name = "global");
 
-  std::unique_ptr<FieldT<Real>> CreateField() const override;
+  OmegaHFieldLayout(Omega_h::Mesh& mesh, Omega_h::Read<Omega_h::I8> mask,
```
jacobmerson: Comment that mask is to be updated with geometric information and point to issue that we can use to track discussion on that project.

@hgangwar please open an issue describing the two modes of mask creation that we discussed yesterday. We can iterate on API design in that issue.

src/pcms/coupler2.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+  PCMS_FUNCTION_TIMER;
+
+  fields_.push_back(std::move(field));
+
+  FieldPtr field_ptr = std::visit(
+    [this, name](auto& field_ptr) -> FieldPtr { return field_ptr.get(); },
+    fields_.back());
+
+  // FieldCommunicator2Ptr field_communicator = std::visit(
```
jacobmerson: Same issue about commented code. Please indicate why this is needed.

src/pcms/coupler2.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+void Application2::AddField(std::string name, OwnedFieldPtr field,
+                            bool participates)
+{
+  PCMS_FUNCTION_TIMER;
+
+  fields_.push_back(std::move(field));
+
+  FieldPtr field_ptr = std::visit(
```
jacobmerson: I don't understand why this is needed, and it seems dangerous. You are doing a visit on a variant to extract out the variant?

src/pcms/coupler2.cpp
```diff
@@ -0,0 +1,70 @@
...
+                                                     *field_ptr);
+    },
+    field_ptr);
+
+  auto [it, inserted] =
+    field_communicators_.insert_or_assign(name, std::move(field_communicator));
+
+  if (!inserted) {
+    std::cerr << "Field with this name" << name << "already exists!\n";
+    std::terminate();
```
jacobmerson: replace with exception

src/pcms/coupler2.cpp
Sichao25: I think the visit here is trying to convert an owned variant to a non owned one.

src/pcms/adapter/meshfields/mesh_fields_adapter2.h
Sichao25: The issue is created. https://github.com/SCOREC/pcms/issues/282

jacobmerson:
@Sichao25 running the GPU tests.

src/pcms/adapter/meshfields/mesh_fields_adapter.h
```diff
@@ -1,5 +1,5 @@
-#ifndef PCMS_COUPLING_OMEGA_H_FIELD_H
-#define PCMS_COUPLING_OMEGA_H_FIELD_H
+#ifndef PCMS_MESH_FIELDS_ADAPTER_H
```
jacobmerson: make this `PCMS_ADAPTER_MESHFIELDS_MESH_FIELDS_ADAPTER_H` (all the include guards are essentially the file path+filename)

src/pcms/adapter/meshfields/mesh_fields_adapter2.h
```diff
@@ -1,37 +1,61 @@
-#include "omega_h_field2.h"
-#include "omega_h_field.h"
-#include "pcms/adapter/omega_h/omega_h_field2.h"
+#ifndef PCMS_MESH_FIELDS_ADAPTER2_H
```
jacobmerson: Make this `PCMS_ADAPTER_MESHFIELDS_MESH_FIELDS_ADAPTER2.h`

src/pcms/adapter/meshfields/mesh_fields_adapter_layout.h
```diff
@@ -1,5 +1,5 @@
-#ifndef PCMS_OMEGA_H_FIELD_LAYOUT_H
-#define PCMS_OMEGA_H_FIELD_LAYOUT_H
+#ifndef PCMS_MESH_FIELDS_ADAPTER_LAYOUT_H
```
jacobmerson: `PCMS_ADAPTER_MESHFIELDS_MESH_FIELDS_ADAPTER_LAYOUT_H`

jacobmerson (approved):
Few minor changes, but it looks good. Go ahead and merge once you update the naming.

jacobmerson (approved):

PR #277, opened 2026-03-08
Update python path handling

### LINKED ISSUE #276: python package not installed properly in spack environment
when `pcms+python` is installed, the python package is not loaded automatically, as is expected based on the behavior with `extends('python')`

jacobmerson:
This PR fixes two issues with the python installation:
1. a `pyproject.toml` file and associated changes to the installation path are added
2. rpath handling and installation location are fixed, so that when `pcms+python` is build in a spack environment, the python package can directly be imported and used.

fixes #276


Sichao25 (commented):
Looks good to me!

PR #209, opened 2025-10-22
Mesh intersection routines 

abhiyanpaudel:
This pull request adds routines for performing Galerkin projection using mesh intersections. The implementation supports linear, 2D triangular elements and enables conservative field transfer between non-matching meshes.

PR #281, opened 2026-03-09
Cleanup mesh intersection

jacobmerson:
This PR is an extension of #209. I refactored Abhiyan's code to get it closer to a state that we can merge.

@abhiyan123 if you would rather merge this and work off of your PR. I'm fine with that, I just want to get the first pass at mesh_intersections in quickly.


jacobmerson:
@Sichao25 you may want to wait until @abhiyan123 indicates if he will merge these changes back to #209 or not to do the review. Once he indicates his preference I'd like to move quickly to get this finalized.

There is still a lot of cleanup that *could* and maybe *should* be done to better decouple components. My strategy was to try to at least:
1. move all non-templated functions out of headers
2. factor out complex code that was repeated multiple times

The error handling bits of this are still a mess, but the PR is already big enough as is. We will have to take a pass at resolving the error handling mismatch when we fix the exceptions in the rest of the library. For the petsc errors, we can utilze the macro/exception type I previously implemented for MuMFiM.

It looks like using PkgConfig for the petsc configuration broke the CI build...In my personal builds I've had better luck with the `PkgConfig` based targets, but we'll need to double check if there is an easy way to fix the CI for this.

jacobmerson:
The issue with the `PkgConfig` based petsc installation here is that the Petsc target only seems to behave nicely when petsc is build in shared library mode.

examples/external-usage-example/CMakeLists.txt
```diff
@@ -6,6 +6,8 @@
 cmake_minimum_required(VERSION 3.19)
 project(pcms_installation_test)
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
```
Fuad-HH, line 9: It should be
```cmake
list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../cmake")
```
with the current directory structure you have. Looks like it is the error CI is showing.

Fuad-HH: However, is it ideal to require it downstream? Looks like you already have petsc added as a dependency in the config file with hints.

jacobmerson: @Fuad-HH I just pushed some updates that resolve the issue (at least on my local system). I had mistakenly copied the `cmake` folder to the wrong directory. I took out the `HINTS` for now.

Longer term, maybe we should install our version of `FindPETSc.cmake` along with the project, then the config file could include the path for the installed `FindPETSc.cmake`. This would avoid the downstream package needing to copy that file.

jacobmerson: Actually better yet, I should just export the `Petsc` target that I create when I configure pcms. That way it will maintain all the paths/link options that are used when PCMS is compiled.

jacobmerson:
@abhiyan123 all but the one test is running on GPU. We just need to debug that and we should be able to merge.

Issue #290, opened 2026-03-15
self-hosted runner build times


jacobmerson:
I maintain build scripts for all of the software dependencies that my students are using including a list of modules, etc. to use [here](https://github.com/laceS-LAB/software-dependencies). I have recently updated this to now match the configured versions of dependencies that we use in CI.

I keep the installed artifacts up to date here: "/uers/mersoj2/laces-software/"

Since all of our dependency versions are locked down, it seems to make a lot of sense to just point our asics CI build to my pre-built dependencies (I can easily add an AMPERE80 build that will work on asics).

This will probably save 80% of the self-hosted CI time and make it more likely that the build versions get updated somewhat consistently.

PR #291, opened 2026-03-15
This PR uses pre-built dependencies for asics runner

jacobmerson:
The current CI workflow is very expensive because it rebuilds lots of dependencies from scratch. To help alleviate this process, this PR makes use of pre-built dependencies (see #290). The risk of this change is if the dependencies do not get maintained on the SCOREC system, or if they become out of date with the rest of the CI infrastructure. However, since multiple folks use these dependencies on SCOREC, they will hopefully be maintained.

@cwsmith @Sichao25 you may want to comment on the pitfalls of this approach.


jacobmerson:
For testing this PR, I had to make `/users/mersoj2` `chmod o+x`. The pre-built stuff should be hosted in `/space/`, or in the `/projects` directory before merging. Not ideal for CI to be able to arbitrarily list someone's home directory.

jacobmerson:
With this change the self-hosted runner completes in 7 minutes instead of somewhere between 48 min - 1 hr. I think we should move forward with this approach as soon as I can move the prebuilt dependencies out of my home directory.

cwsmith:
Nice improvement!  Thank you.

IIRC, we have spack packages for these and could install them alongside the 'normal' user accessible modules instead of treating them as something 'special'.  If we did that, we'd want to periodically install newer versions of the deps (ideally through cron etc.) to catch problems caused by upstream changes.

jacobmerson:
In terms of how we use spack for this. We could create an environment with all the dependencies then do `spack env loads pcms-env` to get a list of relevant modules that can be loaded. Is that what you are thinking?

cwsmith:
I prefer to just expose the modules (i.e., as instructed /opt/scorec/spack/rhel9/README) instead of having users run spack commands.

jacobmerson:
Yes. That's what I meant. That command will give you a list of modules that replicate the spack environment. I didn't mean run spack in CI.

Sichao and I can work on getting this in place when he's back from vacation.

jacobmerson:
@Sichao25 Can you take a look at implementing this strategy on asics again? This significantly reduces the amount of re-build we need to do on asics from the dependencies.

Sichao25:
Sure. It will be a great improvement on self-hosted CI.

jacobmerson:
@Sichao25 it might be worth coming back to this since we are working though so many PCMS PRs currently. This will significantly reduce the amount of time it takes to run the test suite on asics.

jacobmerson:
@Sichao25 looks like there may have been some additional fixes that were needed? Feel free to merge when you are ready.

Sichao25 (commented):
I've added all the fixes. Ready to merge.

Sichao25 (approved):

PR #293, opened 2026-03-16
remove spdlog

### LINKED ISSUE #257: print[Info|Error] ignores setting of stdout/stderr when spdlog is enabled at compile time
The branch here 
https://github.com/SCOREC/pcms/compare/cws/testPrint?expand=1
demonstrates a use case I was hoping `printInfo` would support for debugging.  Specifically, I wanted to create one file per rank and have each rank's calls to `printInfo(...)` write to that file.  

After some digging, I found that when spdlog is enabled, this line 

https://github.com/SCOREC/pcms/blob/bae226d38215d6ee726c1505e56ec99dfec760c9/src/pcms/print.h#L40

ignores what the user set stdout (and likewise for stderr) and writes using spdlog.

This isn't exactly a bug, but it is unexpected behavior.  Ideally, I'd like support for the debugging use case described above and should only require some minor changes.

jacobmerson:
closes #257

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -193,7 +193,6 @@ jobs:
                   -DPETSC_DIR=${{ runner.temp }}/petsc-openmpi \
                   -DPETSC_ARCH=ubuntu-kokkos \
                   -DPCMS_TIMEOUT=10 \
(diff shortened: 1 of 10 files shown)


cwsmith (approved):

hgangwar (commented):
Removes spdlog completely.

PR #294, opened 2026-03-16
enable building without petsc

jacobmerson:
This commit fixes the CMake and test headers, so that PCMS can build without having PETSC installed (albeit without the projection methods)


jacobmerson:
closes #295

cwsmith (approved):

Issue #296, opened 2026-03-16
default `PCMS_ENABLE_PETSC=OFF`


jacobmerson:
@cwsmith brought this up offline, and building without petsc is likely the most sane default since it is really only needed for performing conservative field transfer operations.

PR #298, opened 2026-03-16
default petsc to off

cwsmith:
see #296


jacobmerson (approved):

jacobmerson:
I did double check and the tests run with PCMS_ENABLE_PETSC=ON only. That ensures we test the correctness of the code, but not the build.
@cwsmith I think it would be better if we add a build configuration with `PCMS_ENABLE_PETSC=OFF`.

jacobmerson:
Thanks!

PR #300, opened 2026-03-20
stochastic galerkin projection routines 

abhiyanpaudel:
This PR adds the stochastic Galerkin projection workflow. Currently, this works for 2D linear elements. A follow-up cleanup would be to remove the current file-based Sobol input path and add a function for Sobol sampling.


src/pcms/transfer/calculate_load_vector.cpp
```diff
@@ -5,7 +5,7 @@
 
 namespace pcms
 {
-PetscErrorCode calculateLoadVector(Omega_h::Mesh& target_mesh,
+PetscErrorCode calculateLoadVectorMI(Omega_h::Mesh& target_mesh,
```
Sichao25, line 8: If I haven't missed anything, the `calculateLoadVectorMI` and `calculateLoadVectorMC` share roughly 90% of their code, there may be room to refactor and eliminate the duplication.

src/pcms/transfer/calculate_load_vector.cpp
```diff
@@ -21,7 +21,52 @@ PetscErrorCode calculateLoadVector(Omega_h::Mesh& target_mesh,
 
   // Fill COO values
   auto elmLoadVector =
-    buildLoadVector(target_mesh, source_mesh, intersection, source_values);
+    buildLoadVectorMI(target_mesh, source_mesh, intersection, source_values);
+
+  auto hostElmLoadVector = Kokkos::create_mirror_view(elmLoadVector);
+  Kokkos::deep_copy(hostElmLoadVector, elmLoadVector);
```
Sichao25, line 27: There is a `create_mirror_view_and_copy ` function.

src/pcms/transfer/load_vector_integrator_mc.cpp
```diff
@@ -0,0 +1,243 @@
...
+namespace pcms
+{
+// TODO:: create a function that can sample sobol sequences instead of
+// reading from file
+Kokkos::View<MeshField::Real* [3]> read_sobol_barycentric_samples_from_file(
+  std::string file_path)
+{
+  std::ifstream infile(file_path);
+  if (!infile) {
+    throw std::runtime_error("Could not open sobol sample file");
```
Sichao25, line 13: We already have `pcms_error`.

Sichao25 (commented):
Looks good overall. Just commented a few minor issues.

jacobmerson:
I have a branch I'm working on that integrates this a bit more cleanly into the rest of the updated PCMS framework. I will hopefully have that together for review in the next week or two.

jacobmerson:
I tried to generalize the projection capabilities in #335 and have a MC version of this stochastic projection in place based on this PR's code.

@abhiyan123 I'm going to close this PR for now. Take a look at #335 and think about the structure that's in place there where the integration routines are split into a two phase operation 1) gather points and 2) evaluate integral based on values at gathered points. The goal was to decouple the source evaluations from the integrators (most likely at the cost of some performance).

PR #302, opened 2026-03-21
Extract field creation

jacobmerson:


jacobmerson:
this was all based on an old version of the PCMS api that was subsumed by the new Field/FunctionSpace work.

PR #309, opened 2026-04-07
Fix flat array to rank-2 array conversion

abhiyanpaudel:
Fixes a bug in the conversion from a flattened 1D array representation to a rank-2 array. The previous conversion could produce incorrect indexing/layout when reshaping data from a 1D flattened representation into a rank-2 array. This bug was found while working on PR #300.


jacobmerson:
@abhiyanpaudel are all the changes in this PR needed to fix the bug? I think we should rebase the specific fix maybe [ad5b512](https://github.com/SCOREC/pcms/pull/309/commits/ad5b5121e1277252111c304b35cc3668ef0ca797) on develop.

Wouldn't specifying the Layout of the Kokkos view resolve this without copying?

PR #314, opened 2026-04-20
Fixes a bug in the conversion from a flattened 1D array representation to a rank-2 array.

abhiyanpaudel:
The previous conversion could produce incorrect indexing/layout when reshaping data from a 1D flattened representation into a rank-2 array. This bug was found while working on PR [#300](https://github.com/SCOREC/pcms/pull/300).


jacobmerson:
@Sichao25 Can you confirm that this is resolved in the new branch?

Sichao25:
Yes, we created a new helper function to construct a Kokkos View from an Omega_h array.

jacobmerson:
I believe we addressed this. Please reopen if this comes up again in current version.

PR #317, opened 2026-05-08
Remove kokkos only test case `test_svd_factorization`

Sichao25:
The `test_svd_factorization` test only include Kokkos kernel features and can likely be removed from PCMS. 

For the record, it also has a synchronization issue on Blackwell B120 GPUs, which may be related to the use of team scratch shared memory with size 1 hierarchical parallelism.


Sichao25:
Hey, @abhiyan123 Would you mind confirming that this test is no longer needed anywhere?

jacobmerson (approved):

jacobmerson:
This was used for evaluating Kokkos kernels, but we don't need this anymore since we unit test the function used to compute svd.

Issue #242, opened 2025-12-13
Support for EQDSK fields


jacobmerson:
This will make use of the new gpu-based spline interpolation and can be used to support the Tokamak use-case in https://github.com/SCOREC/STOMMS

PR #318, opened 2026-05-15
Support EQDSK field

Sichao25:
Support the load of EQDSK file and perform evaluation with uniform grid spline interpolator. Followed the data structure implemented in STOMMS. Close issue #242


jacobmerson:
@Sichao25 can you take a look at what the configuration issue is? I got a bit behind on approving PRs while asics was offline.

Sichao25:
@jacobmerson We have two pending updates in self hosted tests #322  and cmake tests #321 .

jacobmerson:
Thanks I had missed those. They are merged now.

jacobmerson:
I think we should add `SplineFunctionSpace::FromEQDSK` that can consume the `EQDSKData` struct. That would remove a lot of the boiler plate that the user would need to deal with when they want to work with EQDSK files. Internally, that could call the uniform grid constructor.

Sichao25:
@jacobmerson The new constructor has been implemented.

src/pcms/field/eqdsk_field.h
```diff
@@ -0,0 +1,77 @@
...
+ * 1. Reads the EQDSK file into an EQDSKData
+ * 2. Creates a SplineFunctionSpace from the data
+ * 3. Creates a Field<Real> on that function space
+ * 4. Populates the field's DOF holder with the PSIZR flux data
+ *
+ * @param filename Path to the EQDSK equilibrium file.
+ * @return EQDSKFieldWithData A struct containing the data, function space,
+ *                            and populated field, ready for evaluation.
+ */
+inline EQDSKFieldWithData MakeEQDSKField(const std::string& filename)
```
jacobmerson, line 67: @usmanriaz07 Are there potentially other fields in the EQDSK that may need to be evaluated later, or is the only field we evaluated in the eqdsk the psi field?

src/pcms/field/eqdsk_field.h
```diff
@@ -0,0 +1,77 @@
...
+#include "pcms/utility/eqdsk.h"
+#include "pcms/field/layout/uniform_grid.h"
+#include "pcms/field/function_space/spline.h"
+
+#include <string>
+
+namespace pcms
+{
+
+struct EQDSKField
```
jacobmerson, line 13: Is there ever a time where we would have an eqdsk with no data associated with it?

src/pcms/field/function_space/spline.cpp
```diff
@@ -27,6 +27,11 @@ SplineFunctionSpace SplineFunctionSpace::FromUniformGrid(
   return SplineFunctionSpace(layout, std::move(evaluator_factory));
 }
 
+SplineFunctionSpace SplineFunctionSpace::FromEQDSK(EQDSKData eqdsk_data)
```
jacobmerson: Take by `const &`

src/pcms/field/function_space/spline.h
```diff
@@ -24,6 +25,8 @@ class SplineFunctionSpace : public FunctionSpace
   [[nodiscard]] static SplineFunctionSpace FromUniformGrid(
     const UniformGrid<2>& grid, CoordinateSystem coordinate_system);
 
+  [[nodiscard]] static SplineFunctionSpace FromEQDSK(EQDSKData eqdsk_data);
```
jacobmerson: I think I was misguided asking for this constructor since this adds coupling between the spline and eqdsk. Since we decided to add the functionality of `MakeEQDSKField` we can just pass the grid to the spline function there.

src/pcms/field/eqdsk_field.h
Sichao25: There will always be an `EQDSKData`, but it doesn't have to be created with the helper functions defined here. It can be constructed manually instead of loaded directly from a file. The struct defined here is just for return values.

src/pcms/field/eqdsk_field.h
```diff
@@ -0,0 +1,77 @@
...
+{
+  SplineFunctionSpace space;
+  Field<Real> field;
+};
+
+struct EQDSKFieldWithData
```
jacobmerson, line 20: Add a comment ``//TODO, @jacobmerson replace with Function<T> . I think this is a good example case for that "Function" interface I am testing out.

src/pcms/field/eqdsk_field.h
jacobmerson, line 67: This function should just create EQDSKData and then call the version of the function that takes the EQDSKData.

usmanriaz07: We do use poloidal current value in the field line tracing. See the following piece of code:
https://github.com/SCOREC/STOMMS/blob/main/stomms_meshgen/src/eqdskData.cc#L197-L203

In ezfit.F90, the relevant function is: https://github.com/SCOREC/STOMMS/blob/main/stomms_meshgen/src/ezfit.F90#L562-L578

Other than poloidal current, we only need psi field.

Sichao25: They are included as I_psi and psi_grid in `EQDSKData`.

jacobmerson (approved):

PR #308, opened 2026-04-02
Remove `fuzz` from Point Search, Return Face ID alongside Element ID

matthew-mccall:
This pull request updates the mesh field adapters and point search logic to consistently use `face_idx` instead of `elem_idx` for identifying mesh elements, and removes the use of the `fuzz` parameter from 2D triangle-bounding box intersection logic. The changes improve clarity and correctness in how mesh faces are referenced and streamline the intersection checking code.

**Mesh element indexing consistency:**
- Updated all relevant functors and evaluation routines in `mesh_fields_adapter.h` and `mesh_fields_adapter2.h` to use `face_idx` instead of `elem_idx` when referencing mesh faces, ensuring consistent and correct indexing throughout the codebase. [[1]](diffhunk://#diff-af689a586c4a51500ee0efc8af6a514502403c4de7a7029aae4b3ab841c11470L350-R354) [[2]](diffhunk://#diff-af689a586c4a51500ee0efc8af6a514502403c4de7a7029aae4b3ab841c11470L388-R392) [[3]](diffhunk://#diff-7a26dc3a69ac86f6397f4334f08d5df526ffcbe1b5a25f1410fdcf21e6891ce4L136-R137) [[4]](diffhunk://#diff-7a26dc3a69ac86f6397f4334f08d5df526ffcbe1b5a25f1410fdcf21e6891ce4L169-R176) [[5]](diffhunk://#diff-7a26dc3a69ac86f6397f4334f08d5df526ffcbe1b5a25f1410fdcf21e6891ce4R247-R250) [[6]](diffhunk://#diff-7a26dc3a69ac86f6397f4334f08d5df526ffcbe1b5a25f1410fdcf21e6891ce4L262-R267)

**Point search and intersection logic simplification:**
[...]

PR #319, opened 2026-05-15
Remove fuzz and return face id alongside element id (Replace stale #308)

Sichao25:
Resolved the merge conflict and updated the PR #308  to fit the new field API.


src/pcms/localization/point_search.h
```diff
@@ -60,6 +124,19 @@ class PointLocalizationSearch
...
+   * This function provides a temporary solution to retrieve the original
+   * behavior of the point search, which previously returned the face ID
+   * regardless of which entity the search result belonged to. Many parts of
+   * the codebase still rely on this legacy behavior. After updating the point
+   * search to return the exact entity, this function can be called to retrieve
+   * the old behavior and ensure correctness. Long term, the implementation
+   * should be updated to properly handle the new result.
+   */
+  [[nodiscard]] virtual LO GetOwningElementId(const Result& result) const = 0;
+  [[nodiscard]] virtual Kokkos::View<LO*> GetOwningElementIds(
```
jacobmerson, line 111: Can these GetOwningElementId functions be made into free functions?

Sichao25: It is possible, but since we need to access some private members, more API changes such as getter functions are required.

jacobmerson: Ok, so I think we should probably have a free function that given `mesh`, `dim`, `id` gives back the owning element. The member function can call this.

The free function version will be useful in a number of other places.

src/pcms/localization/point_search.h
```diff
@@ -22,12 +60,76 @@ namespace detail
...
+ * (triangle) by computing the triangle area from two edge vectors using the
+ * parallelogram formula valid in any dimension:
+ *   area = 0.5 * sqrt(||a||^2 ||b||^2 - (a·b)^2)
+ *
+ * Template parameter N is the embedding dimension (e.g., 2 for planar, 3 for
+ * spatial), while the simplex is always a triangle (3 vertices).
+ */
+template <int N>
+[[nodiscard]] KOKKOS_FUNCTION inline Real
+distance_to_closest_edge_from_barycentric(const Omega_h::Matrix<N, 3>& coords,
```
HelenW42: In the case when a point is on the infinite extension of one of the face's edges, this function will erroneously say the point is 0 units from an edge. Also, this method will not (as the comment states) extend to 3D: barycentric coordinates in 3D are the ratios of volumes of tetrahedra, thus the product of the ith coordinate and the height/altitude of the face is not in the distance of the point to the edge.

jacobmerson: @HelenW42 did you end up having a fix for this function?

Sichao25: I am working on a fix in 2D that computes the distance to the three edges and compares them, while keeping the original function as an efficient solution for points inside the element. I haven't implemented or thought much about the 3D case yet. Happy to hear any thoughts or if there are any known solutions.

HelenW42: The attached PDF has my (rough) derivations of how to find whether a point is within a certain tolerance of an edge, vertex, and face in 2D and 3D from the barycentric coordinates alone. While I didn't include a specific fix for the collinear edge case, it's not that expensive an addition: if the point seems to have zero distance from the edge and is not within it, just check the distance from the vertices that define that edge (the derivation for which is in the file).
[bary_to_global_tolerance.pdf](https://github.com/user-attachments/files/29868937/bary_to_global_tolerance.pdf)

Sichao25: Based on an offline discussion with Jacob, it should be okay to remove the function `distance_to_closest_edge_from_barycentric` from this PR since the current `GridPointSearch` class doesn't use it during the search process. I'll add a test case covering the edge case for the current approach. We can then address this barycentric-based approach in a separate issue or PR.

Sichao25:
@jacobmerson  The point search function based on barycentric coordinates without reference has been removed.

jacobmerson (approved):

PR #321, opened 2026-06-03
Upgrade petsc to 3.25.2

Sichao25:
PETSc 3.24.2 incorrectly reads C++17 from the Kokkos cmake config when Kokkos is compiled with C++20. This causes compilation errors when building code that depends on C++20 features. It is unclear why this did not cause issues previously, but upgrading to PETSc 3.25.2 resolves the problem.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -133,7 +133,7 @@ jobs:
       id: clone-petsc
       run: |
         cd ${{ runner.temp }}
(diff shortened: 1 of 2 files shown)


jacobmerson:
rerunning now that asics is back online.

jacobmerson (approved):

jacobmerson:
Approve with self hosted runner failing since this is one of two issues that needs to be resolved to get the asics runner working...

PR #322, opened 2026-06-16
Upgrade adios2 to latest in self hosted test

Sichao25:
Upgrade adios2 to include latest cleanup on some unused cmake installations.


Sichao25:
@jacobmerson Since the only change here is in self hosted test, it makes sense to me to merge. The error in other tests have been handled in #321

jacobmerson (approved):

PR #329, opened 2026-07-07
Add missing vtk functions/parameters in python API

Sichao25:
Added the missing essential functions and parameters to the Python VTK I/O API, along with tests updated.


jacobmerson:
@Sichao25 and I had an offline discussion. We decided that it makes more sense to remove these omega_h python readers from PCMS and will instead work towards utilizing the updated Omega_h APIs directly.

PR #330, opened 2026-07-09
Replace omega_h functions in Python API with PyOmega_h

Sichao25:
Since Omega_h has updated its python api, it is safe to remove the redundant Python bindings from PCMS. Note that this change will require the related libraries to be built as shared libraries.


jacobmerson:
-908 lines of code...I'm a fan. Once this is done it closes #329

jacobmerson:
Is there a specific version numer of MPI that will be required for this? We should bump the minor version in CMake.

We should also update the spack package (or at least open an issue there mentioning the new omega_h version requirements)

https://github.com/jacobmerson/pcms-spack

Sichao25:
Do you mean the Omega_h version? I'm not aware of this PR changing any MPI version requirements.

jacobmerson:
Sorry I meant Omega_h...that's what 4 hours of sleep does to your brain.

jacobmerson (approved):

Issue #331, opened 2026-07-10
API array inconsistency


jacobmerson:
The APIs should be consistent enough that we can avoid copying data everywhere. One such problematic mismatch is between the point evaluation routines and the SetDOFData routines.

See here: https://github.com/SCOREC/pcms/blob/5697725e47294d2a41b093821fe07f0122a37176/src/pcms/transfer/interpolator.h#L53-L69

In the short-term with scalar fields, we can just create a 1D view that aliases the 2D view. However, longer term, the SetDOFData API needs to be updated to follow the cannonical data format we have been using in this project.

i.e., data is always Rank 2 where the shape is [# npoints][# components]

jacobmerson:
Note this also impacts `GetDofHolderData`, we should also be returning rank2 arrays.

PR #332, opened 2026-07-11
API cleanup to use 2D arrays for Set/GetDOFs

jacobmerson:


jacobmerson:
This fixes #331

jacobmerson:
Replaced by #339. Rebased onto current develop; head moved to the SCOREC-side `jacobmerson/set-get-2d-arrays` branch so it can serve as the stacking base for the rest of the stack.

PR #334, opened 2026-07-11
style: clang-format linear_interpolant.hpp and adj_search.hpp

jacobmerson:
reformat headers with clang-format-18


jacobmerson:
This popped up from a mismatch between my local clang-format and the one in CI. Apparently behavior for 18.1.8 is different from 18.1.3 that's used in ubuntu...

PR #335, opened 2026-07-12
Conservative field transfer (integrator + Monte-Carlo projection)

jacobmerson:
Adds conservative field-transfer support built on the integrator-based (Galerkin) and Monte-Carlo / control-variate projection operators, replacing the old `calculate_mass_matrix` / `load_vector` implementation.

Highlights:
- Integrator-based conservative (Galerkin) projection transfer
- Monte-Carlo / control-variate conservative projection
- DOF holders indexed by a compact GID permutation; coupler separates GID-layout and field messages
- Python bindings expose conservative transfer

Base is `jacobmerson/various-fixes` (#340 ).


jacobmerson:
There are a few GPU issues I'm still sorting out.

jacobmerson:
GPU builds are now working/tests passing (at least on othello)

src/pcms/transfer/integration_point_set.hpp
```diff
@@ -0,0 +1,57 @@
...
+// be evaluated for integration. The ordering is significant: sampled values
+// passed back to a LinearFormIntegrator must use the same point order as
+// stored here.
+//
+// This type intentionally does not expose quadrature weights, basis values,
+// entity ids, or any other assembly metadata. Those remain private to the
+// concrete integrator implementation. It owns the coordinate buffer; callers
+// receive a non-owning CoordinateView from GetCoordinates().
+template <typename MemorySpace>
+class IntegrationPointSet
```
Sichao25: This looks like a thin wrapper that just owns the coordinates and provides a coordinate view. Do we need this instead of just a plain Kokkos view?

src/pcms/coupler/serializer/xgc.h
```diff
@@ -41,7 +41,9 @@ class XGCFieldSerializer : public FieldSerializer<T>
       const LO num_dof = static_cast<LO>(data.extent(0));
       const LO num_comp = static_cast<LO>(data.extent(1));
       for (LO i = 0; i < num_dof; ++i) {
-        if (owned[i]) {
+        // A negative permutation entry marks a holder outside the exchange
+        // (non-owned, or owned but outside the overlap region); it has no slot.
+        if (owned[i] && permutation[i] >= 0) {
```
Sichao25: The permutation is constructed when an holder is owned and in the overlap domain. If so, does the permutation itself already include the 'owned' info? Is there a use case where we'd need to check both?

src/pcms/transfer/omega_h_form_integrator_utils.hpp
```diff
@@ -0,0 +1,341 @@
...
+      tri_coords[2] = {p2[0], p2[1]};
+
+      Omega_h::Few<Omega_h::Vector<2>, 2> basis;
+      basis[0] = tri_coords[1] - tri_coords[0];
+      basis[1] = tri_coords[2] - tri_coords[0];
+
+      Omega_h::Real area =
+        Kokkos::fabs(Omega_h::triangle_area_from_basis(basis));
+
+      const double eps_area = abs_tol + rel_tol * poly_area;
```
Sichao25, line 229: This might be irrelevant to the implementation here, but I'm wondering how `abs_tol` and `rel_tol` are determined? Is there a rule for this? I want to double-check because I remember a mismatch between the fuzz value （already be removed from point search and epsilon from meshfields causing a bug in some cases.

Sichao25 (commented):
Overall looks good to me.

Sichao25:
The `OmegaHMassIntegrator` test failed. I will take a look.

src/pcms/transfer/integration_point_set.hpp
jacobmerson: Yeah, we can drop this. I originally was thinking of using that as an abstract base, but the use-case I had in my mind was more theoretical than executed in the codebase.

src/pcms/coupler/serializer/xgc.h
jacobmerson: I believe that's correct. with the current version of the permutation, we explicitly have negative values for any non-owned entities. We probably need to formalize this somehow. At least through comments, but we may want to add an assert in the loop and add test cases that check that the serialized array is negative for non-owning entries.

src/pcms/transfer/omega_h_form_integrator_utils.hpp
jacobmerson: At the moment, I believe these are hard-coded, but this code will eventually need to be made more robust. @abhiyan123 may have more details on why those triangles are skipped, but I'm pretty sure it was to avoid degeneracies that can happen with slivers, etc.

Long term, we should look into simulation of simplicity https://dl.acm.org/doi/10.1145/77635.77639 , or other methods that can help to avoid these cases without tolerances.

Sichao25:
I've added a few commits addressing the test failures and the comments. If it looks good, it's ready to merge.

jacobmerson:
@Sichao25  Thanks!

PR #336, opened 2026-07-12
Field/Function split: FieldFactory + two-level naming, remove AddLayout

jacobmerson:
Splits field construction from the function-space concept and removes `AddLayout`, reworking field construction and coupler registration around a two-level naming model.

Highlights:
- Update Field/Function type hierarchy
- Introduce `FieldFactory`; rework construction around two-level naming, removing `AddLayout`
- Fix the `CreateFunction` / `CreateField` split

Base is `jacobmerson/conservative-transfer` (#335)


Sichao25 (commented):
Looks good to me.

PR #341, opened 2026-07-13
Add global comm

hgangwar:
Moving PR #228 to here for a clean PR.

This PR introduces `redev::AdiosGlobalComm` to provide a global communication interface between the coupler and the applications using the existing ADIOS channel.
**Usage:**
```cpp
auto GDI = app->AddData<pcms::GO>("mean", mean, comm);
````
This creates a global communicator that can send and receive data, for example:
```cpp
app->BeginSendPhase();
app->SendData("mean");
app->EndSendPhase();

app->BeginReceivePhase();
app->ReceiveData("mean");
app->EndReceivePhase();
```
- The `test_GDI` test case verifies global communication and data transfer.


jacobmerson:
@hgangwar did you update this based on the redev merge? Note, you will need to make sure you bump the redev version in the config files for the CI.

jacobmerson:
@hgangwar I think we discussed having this owned by the application so we have a consistent API between fields and "global data"

```
AddField -> AddData
SendField -> SendData
ReceiveField -> ReceiveData
```

jacobmerson (changes requested):
See previous comment about making the API match between fields and global data.

jacobmerson (commented):
Sorry I did not read carefully, you did add the global data communicator to the application. However for consistency:
- `SendData` should be a function on the application that takes a string
- `Send` should be a function on the GDI pointer.

jacobmerson:
Also, look at how the `FieldHandle` works. I.e., we return a cheap to copy handle of the field which stores a string which is used to actually perform the send/receive. That way we are not handing a pointer back to the user.

src/pcms/coupler/coupler.hpp
```diff
@@ -31,6 +65,28 @@ class FieldHandle
...
+  DataHandle(Application* app, std::string name)
+    : app_(app), name_(std::move(name))
+  {
+  }
+
+  void Send(T* msg, std::string variable_name, std::size_t msg_size,
+            redev::Mode mode = redev::Mode::Synchronous) const;
+
+  [[nodiscard]] std::vector<T> Receive(
+    std::string variable_name, std::size_t msg_size,
```
jacobmerson: Shouldn't this API be consistent with `Send` i.e., you should pass in the appropriate `array_view` which gets filled up. otherwise, a buffer needs to be created on the heap for every call to recieve.

src/pcms/coupler/coupler.hpp
```diff
@@ -31,6 +65,28 @@ class FieldHandle
...
+template <typename T>
+class DataHandle
+{
+public:
+  DataHandle(Application* app, std::string name)
+    : app_(app), name_(std::move(name))
+  {
+  }
+
+  void Send(T* msg, std::string variable_name, std::size_t msg_size,
```
jacobmerson: This should take an array_view instead of a pointer and a size.

src/pcms/coupler/coupler.hpp
```diff
@@ -70,7 +126,11 @@ class Application
   FieldHandle<T> AddField(std::string name, Field<T>&& field,
                           std::unique_ptr<FieldSerializer<T>> serializer,
                           bool participates = true);
-
+  template <typename T>
+  DataHandle<T> AddData(std::string name, MPI_Comm mpi_comm);
+  template <typename T>
+  [[nodiscard]] GlobalDataInterface<T>& GetDataInterface(
```
jacobmerson: Should this return the handle we discussed?

jacobmerson (commented):
Don't have time to finish this review right now...but I want to go over again why we need to pass the data directly into `SendData` I understand at the redev level this is needed for the communictor because otherwise we will be constructing a lot of adios2 objects. However, it's less clear why that's needed here. It is possible that it's the correct choice, just the mismatch to how the fields are handled is giving me pause.

You may want to discuss with Sichao or Cameron or Fuad. Show them the general API and see which direction seems more natural to them.

src/pcms/coupler/coupler.hpp
hgangwar: Made the changes, can you please take a look again?

src/pcms/coupler/coupler.hpp
hgangwar: Same as above.

src/pcms/coupler/coupler.hpp
hgangwar: `AddData` is returning a `DataHandle`, this is for `DataHandle` to be able return `GlobalDataInterface` object ptr. This is non-essential for now, so I have removed it.

test/test_GDI.cpp
```diff
@@ -0,0 +1,143 @@
...
+  printf("final Mean = %ld\n", mean[0]);
+  assert(std::fabs(mean[0] - 1.0) < 1e-12);
+  printf("GDI test successful.\n");
+}
+void xgc_total_f(MPI_Comm comm)
+{
+  pcms::Coupler coupler("proxy_couple", comm, false, {});
+  pcms::Application* app = coupler.AddApplication("proxy_couple_xgc_total_f");
+
+  auto GDI = app->AddData<pcms::GO>("global_comm", comm);
```
Fuad-HH: I think writing the returned typename (`DataHandle`) explicitly will make the code more readable.

test/test_GDI.cpp
```diff
@@ -0,0 +1,143 @@
...
+  Kokkos::View<long*, pcms::HostMemorySpace> mean_buffer("mean_buffer", 1);
+
+  pcms::Rank1View<long, pcms::HostMemorySpace> mean{mean_buffer.data(),
+                                                    mean_buffer.extent(0)};
+
+  do {
+    for (int i = 0; i < COMM_ROUNDS; ++i) {
+      app->BeginReceivePhase();
+      GDI.Receive(mean, "mean");
```
Fuad-HH: Like @jacobmerson said, I would also feel more comfortable if the interface was more parallel to field transfer interfaces. For example, here:
1. Instead of `Receive`, `ReceiveData` would be better.
2. This `Receive` method is called on the `DataHandle` object whereas the `ReceiveField` is called on the app. This would look confusing to me as an user.

On the other hand, I like that the data (`mean`) is passed during the `Receive` call. To me, I would like to use one `DataHandle` to transfer many data arrays if needed since there's nothing to cache here, unlike fields. However, this way of passing data makes the `AddData` call a bit confusing because it is not actually "adding" a data.

test/test_GDI.cpp
```diff
@@ -0,0 +1,143 @@
...
+      app->EndSendPhase();
+      printf("delta Sent mean:%ld\n", mean[0]);
+      app->BeginReceivePhase();
+      gdi.Receive(mean, "mean");
+      app->EndReceivePhase();
+      mean[0] = mean[0] / 2;
+    }
+  } while (!done);
+  printf("final Mean = %ld\n", mean[0]);
+  assert(std::fabs(mean[0] - 1.0) < 1e-12);
```
Fuad-HH, line 36: `assert` only works when it's compiled in Debug mode (see [here](DataHandle)). You can use if statement and `std::cerr` like other tests.
I guess `catch2` is also added in pcms. You can use that too. But I wouldn't include it for a single assertion here.

Fuad-HH (commented):
To me, keeping the interface parallel to field transfer APIs is important as it will help user understand the structure easily, specially the highest level calls: field send/receives are called on the application whereas here they are called on the objects returned by `AddData`.

src/pcms/coupler/coupler.hpp
```diff
@@ -11,10 +11,57 @@
...
+  {
+    PCMS_FUNCTION_TIMER;
+
+    auto received = comm_.Receive(
+      variable_name_, static_cast<std::size_t>(data_.extent(0)), mode);
+
+    PCMS_ALWAYS_ASSERT(received.size() ==
+                       static_cast<std::size_t>(data_.extent(0)));
+
+    std::copy(received.begin(), received.end(), data_.data_handle());
```
jacobmerson: @hgangwar I though we updated the redev API to take a ptr/size instead of returning a vector. If not, we need to go back and update that. We should not need to allocate/copy on every receive.

src/pcms/coupler/coupler.hpp
```diff
@@ -98,6 +160,18 @@ class Application
 
   // Register a transferable field: like AddField but retains the function space
   // (via the stored Function) and returns a FunctionHandle usable in transfers.
+  template <typename T>
+  FieldHandle<T> AddField(std::string name, Field<T>&& field,
+                          std::unique_ptr<FieldSerializer<T>> serializer,
+                          bool participates = true);
+  template <typename T>
+  DataHandle<T> AddData(std::string name,
+                        Rank1View<T, pcms::HostMemorySpace> data,
```
jacobmerson: Here when we `AddData` we are pass-by-reference / store ptr to the data. It seems like fields are moved in so the coupler takes owenership of those?

At a minimum, we need to document our strategy here.

src/pcms/coupler/coupler.hpp
```diff
@@ -98,6 +160,18 @@ class Application
...
+                        MPI_Comm mpi_comm);
+  template <typename T>
+  [[nodiscard]] GlobalDataInterface<T>& GetDataInterface(
```
jacobmerson: Why is this needed in the public API? I think you should only allow the user to grab a `DataHandle`.

test/CMakeLists.txt
```diff
@@ -380,7 +411,6 @@ if(Catch2_FOUND)
       APPEND
       PCMS_UNIT_TEST_SOURCES
       test_error_handling.cpp
-      test_eqdsk.cpp
```
jacobmerson, line 383: why are you removing this test case?

src/pcms/coupler/coupler.hpp
hgangwar: Redev receive still returns a vector. I will make a PR for it in redev.

src/pcms/coupler/coupler.hpp
hgangwar: Changed it to pass-by-reference, fixing the input data type to be vector<T>. Now AddData would make a view from the passed reference.

src/pcms/coupler/coupler.hpp
hgangwar: I removed the member function.

test/CMakeLists.txt
hgangwar: Added it back, possibly came in from a merge.

src/pcms/coupler/global_communicator.h
```diff
@@ -32,14 +32,13 @@ struct GlobalCommunicator
     comm_.SetCommParams(VarName, msg_size);
     comm_.Send(msg, mode);
   }
-  std::vector<T> Receive(std::string VarName, size_t msg_size,
-                         Mode mode = Mode::Synchronous)
+  void Receive(T* destination, std::string VarName, size_t msg_size,
```
jacobmerson: In PCMS we have access to `Rank1View`. Please use that here instead of pointer and size.

src/pcms/coupler/coupler.hpp
```diff
@@ -99,6 +155,16 @@ class Application
...
   // (via the stored Function) and returns a FunctionHandle usable in transfers.
   template <typename T>
+  FieldHandle<T> AddField(std::string name, Field<T>&& field,
+                          std::unique_ptr<FieldSerializer<T>> serializer,
+                          bool participates = true);
+  // Registers a reference to user-owned data. The Application does not take
+  // ownership; it uses the registered reference for subsequent send and receive
+  // operations.
+  template <typename T>
+  DataHandle<T> AddData(std::string name, std::vector<T>& data,
```
jacobmerson: If we are using ptr semantics over the array, let's take `Rank1View<HostMemory>` here. That way if someone has an array that is not a vector the function can work with that as well.

jacobmerson (commented):
A few more things to update, but I think this is close to ready.

PR #345, opened 2026-07-28
Added BVH accelerated point localization and updated localization API

HelenW42:
This PR uses `ArborX::BVH` to provide accelerated point localization and refactors the previous localization API (`pcms::PointLocalizationSearch`) to no longer use templates.

### DIFF
--- a/src/pcms/field/evaluator/omega_h_lagrange.h
+++ b/src/pcms/field/evaluator/omega_h_lagrange.h
@@ -65,13 +65,11 @@ struct CopyCoordsFunctor
 template <int Dim>
 OmegaHLagrangeLocHint BuildLagrangeLocHint(
   Omega_h::Mesh& mesh, int mesh_dim,
-  Kokkos::View<typename PointLocalizationSearch<Dim>::Result*,
-               DeviceMemorySpace>
-    results,
+  PointSearch::Results& results,
   Kokkos::View<Real**, DeviceMemorySpace> coords_d,
   Kokkos::View<LO*, DeviceMemorySpace> owning_ids, OutOfBoundsMode mode)
 {
-  LO n = static_cast<LO>(results.size());
+  LO n = static_cast<LO>(results.dimensionalities.size());
 
   // First pass: count valid and missing
   Kokkos::View<int*, DeviceMemorySpace> is_valid("is_valid", n);
@@ -79,7 +77,7 @@ OmegaHLagrangeLocHint BuildLagrangeLocHint(
     "CheckValidity",
     Kokkos::RangePolicy<typename DeviceMemorySpace::execution_space>(0, n),
     KOKKOS_LAMBDA(LO i) {
-      bool out = (owning_ids(i) < 0) || (results(i).element_id < 0);
+      bool out = (owning_ids(i) < 0) || (results.element_ids(i) < 0);
       is_valid(i) = out ? 0 : 1;
     });
 
@@ -344,7 +342,7 @@ public:
         detail::CopyCoordsFunctor<Dim> copy_functor(coords_d, raw_coords);
         Kokkos::parallel_for("copy_coords", n_pts, copy_functor);
 
-        auto results_d = search(coords_d);
+        auto results_d = search.apply(coords);
         auto owning_ids = search.GetOwningElementIds(results_d);
 
         return detail::BuildLagrangeLocHint<Dim>(
--- a/src/pcms/localization/CMakeLists.txt
+++ b/src/pcms/localization/CMakeLists.txt
@@ -7,6 +7,7 @@ set(
         localization_factory.h
         point_cloud_localization.h
         mesh_localization.h
+        point_localization.h
 )
 
 set(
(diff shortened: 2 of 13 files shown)


src/pcms/localization/point_localization.h
```diff
@@ -0,0 +1,479 @@
+#ifndef POINT_LOCALIZATION_H
+#define POINT_LOCALIZATION_H
+
+#include <any>
+
+#include <ArborX.hpp>
```
jacobmerson, line 6: We need to make sure that we can build without ArborX. This will require adding a configure-time option `PCMS_ENABLE_ARBORX` to the `CMakeLists.txt` as well as an entry in `configuration.h.in`.

src/pcms/localization/point_localization.h
```diff
@@ -0,0 +1,479 @@
...
+#include <ArborX_Triangle.hpp>
+#include <detail/ArborX_PairValueIndex.hpp>
+#include <detail/ArborX_AttachIndices.hpp>
```
jacobmerson, line 9: We need to carefully review if APIs from `detail` are needed. Typically things in `detail` files or namespaces are internal implementation details and liable to change without notice.

src/pcms/localization/point_localization.h
```diff
@@ -0,0 +1,479 @@
...
+
+#include <Kokkos_Core.hpp>
+#include <Omega_h_mesh.hpp>
+#include <Omega_h_bbox.hpp>
+#include <Omega_h_shape.hpp>
+#include <Omega_h_matrix.hpp>
+#include <Omega_h_simplex.hpp>
```
jacobmerson, line 16: Make sure we only include these if PCMS is built with `Omega_h`. I.e., use `#IFDEF PCMS_ENABLE_OMEGA_H`

src/pcms/localization/point_localization.h
```diff
@@ -0,0 +1,479 @@
...
+#include "pcms/utility/arrays.h"
+#include "pcms/localization/point_search.h"
+
+namespace pcms
+{
+
+namespace detail
+{
+
+/**
```
jacobmerson, line 29: Is this needed in the header, or can it be moved to the .cpp file?

src/pcms/localization/point_localization.h
```diff
@@ -0,0 +1,479 @@
...
+*/
+KOKKOS_INLINE_FUNCTION
+LO kronecker(LO i, LO j) { return (LO)(i==j); }
+
+/**
+* @brief Omega_h::Mesh tagged with the templated dimension required for 
+*		 compatability with ArborX::BVH
+*/
+template <int dim>
+struct Omega_h_Mesh_Adapt
```
jacobmerson, line 44: Can this be moved to the CPP file? It seems like an implementation detail we don't want to expose in the header.

src/pcms/localization/point_localization.h
jacobmerson, line 1: 1. Rather than commenting on every instance I think most of what's here can be moved to the `.cpp` file since it is specific template instantiations that are only used to drive the implementation of the localization class.
2. We are *slowly* moving all header files over to `.hpp` since this is a new file, let's use the `.hpp` extension.

src/pcms/localization/point_search.h
jacobmerson, line 1: I'm a bit confused about the current file naming (I understand that much of this was inherited). I think we should roughly have the following:

- `localization.hpp`: include base class pure virtual interface for localization
- `uniform_grid_localization.hpp`: class definition for uniform grid-based localization inheriting from the base class.
- `arbor_x_localization.hpp`: class definition for ArborX
- associated .cpp files

We should ameks sure that as much code as possible is in the `.cpp` file rather than the header file. This a) makes it easier for the consumer to know how to interact with the library, b) limits the likelyhood of conflicts from consumers c) increases the compile performance

src/pcms/field/evaluator/omega_h_lagrange.h
```diff
@@ -344,7 +342,7 @@ class OmegaHLagrangeEvaluatorFactory : public FieldEvaluatorFactory<T>
         detail::CopyCoordsFunctor<Dim> copy_functor(coords_d, raw_coords);
         Kokkos::parallel_for("copy_coords", n_pts, copy_functor);
 
-        auto results_d = search(coords_d);
+        auto results_d = search.apply(coords);
```
jacobmerson, line 345: Make sure any methods have PascalCase (capitalize the first letter of every word).

src/pcms/localization/CMakeLists.txt
```diff
@@ -15,7 +16,9 @@ set(
         adj_search.cpp
         mls_support_helpers.cpp
         mesh_localization.cpp
+        point_localization.cpp
 )
+find_package(ArborX REQUIRED)
```
jacobmerson, line 21: Only require this if `PCMS_ENABLE_ARBORX=ON`

jacobmerson (commented):
A few initial things to work on. I will return to this in a little while with some additional feedback.

PR #346, opened 2026-07-29
Debug meshFields backend to fit 1.0.0 API

Sichao25:
Change API to fit a few breaking changes introduced by MeshFields 1.0.0. The return value of `CreateLagrangeField` is no longer the field itself, and the redundant barycentric coordinates have been removed.

### DIFF
--- a/src/pcms/field/evaluator/mesh_fields_backend.h
+++ b/src/pcms/field/evaluator/mesh_fields_backend.h
@@ -425,12 +425,8 @@ struct FillCoordinatesDeviceFunctor
     for (int j = 0; j < (dim_ + 1); ++j)
       coordinates_(index, j) = local[j];
 #else
-    // Computation in barycentric_from_global sometimes produce values
-    // slightly outside [0, 1] for points on or near element boundaries.
     for (int j = 0; j < dim_; ++j) {
-      coordinates_(index, j) = local[j] < 0.0   ? 0.0
-                               : local[j] > 1.0 ? 1.0
-                                                : local[j];
+      coordinates_(index, j) = local[j];
     }
 #endif
     indices_(index) = orig_idx;
@@ -573,12 +569,8 @@ struct MeshFieldsAdapter2LocalizationHint
       for (int j = 0; j < (mesh.dim() + 1); ++j)
         coordinates_(index, j) = local[j];
 #else
-      // Computation in barycentric_from_global sometimes produce values
-      // slightly outside [0, 1] for points on or near element boundaries.
       for (int j = 0; j < mesh.dim(); ++j) {
-        coordinates_(index, j) = local[j] < 0.0   ? 0.0
-                                 : local[j] > 1.0 ? 1.0
-                                                  : local[j];
+        coordinates_(index, j) = local[j];
       }
 #endif
       indices_(index) = static_cast<LO>(orig_idx);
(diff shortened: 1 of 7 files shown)


src/pcms/field/evaluator/mesh_fields_backend.h
```diff
@@ -404,8 +421,18 @@ struct FillCoordinatesDeviceFunctor
                                    global_coords_(orig_idx, 1)};
     const auto local =
       Omega_h::barycentric_from_global<2, 2>(point, vertex_coords);
+#if MeshFields_VERSION < 10000
     for (int j = 0; j < (dim_ + 1); ++j)
       coordinates_(index, j) = local[j];
+#else
+    // Computation in barycentric_from_global sometimes produce values
+    // slightly outside [0, 1] for points on or near element boundaries.
+    for (int j = 0; j < dim_; ++j) {
```
jacobmerson, line 407: @cwsmith @Joshua-Kloepfer This code is required due to the assert statements in MeshFields. I'm wondering if those asserts should have tolerances to allow for floating point computations, or if there is a better way to handle this.

cwsmith: We can relax these checks:

https://github.com/SCOREC/meshFields/blob/47336ac5048f8ba7a45f8a85899006b9fb74a592/src/MeshField_Shape.hpp#L26-L29

https://github.com/SCOREC/meshFields/blob/47336ac5048f8ba7a45f8a85899006b9fb74a592/src/MeshField_Shape.hpp#L53-L56

to use `Epsilon` (`1e-12`) instead of `MachinePrecision` (`1e-15`) defined here:

https://github.com/SCOREC/meshFields/blob/47336ac5048f8ba7a45f8a85899006b9fb74a592/src/MeshField_Defines.hpp#L19-L20

Both seem somewhat arbitrary...

cwsmith: A PR with a looser tolerance is here: https://github.com/SCOREC/meshFields/pull/97

Sichao25: I can remove the check here in pcms, but this seems to create a few incompatible versions between meshfields 1.0.0 and latest. Would that be a problem? If so, we'd probably need a patch on the meshfields side.

cwsmith:
Afaik, we don't need to support old versions of meshfields.

Sichao25:
With the tolerance update in meshFields, the assert clamping values into [0,1] has been removed. I've kept the `#if MeshFields_VERSION` check since some CI and existing workflows may still rely on it. We can remove this check in the future, as there's no need to keep supporting old meshFields. I think this PR is ready for review and merge.

jacobmerson:
@Sichao25 we may want to put the following into a separate PR, but my understanding is that MeshFields now include parametric coordinates of each DOFHolder. These should be used to construct global coordinates in the layout rather than the hard-coding we do for linear/quadratic currently. This should let us more seemlessly support alternative element types (e.g., reduced quintic) and higher degrees.

Sichao25:
Okay. I’ll put that in a separate PR. I think the error here is related to the MeshFields version in the pre-built dependency.  @jacobmerson When you have a chance, would you mind checking whether we’re using a MeshFields version newer than 1.0.0?

jacobmerson:
The prebuilt dependencies are on `b1482bbba288df210784b2345eae08e34faabdc4` what hash should be used? I should update that in my build scripts: https://github.com/LACES-LAB/software-dependencies/blob/main/configure-and-build-meshFields.sh

jacobmerson:
@hgangwar @Fuad-HH @abhiyan123 this PR will be a breaking change when we merge it. I will update my dependency build folder at the same time that merge this PR.

PR #348, opened 2026-08-02
Fix std::array Bug for GPU kernels - For Conservative Transfer

Fuad-HH:
`std::array` in several structs are introducing silent bugs. In this commit, I focused mainly on conservative field transfer related bugs. I have found that conservative projections work with Kokkos _OpenMP_ backend but fails with Kokkos _GPU/CUDA_ backend.

There are structs, for example, UniformGrid, which uses std::array data members which are failing silently inside gpu kernels when these are used as, for example `Kokkos::View<UniformGrid [1]>`.

I visually verified the result using the example conservative round-trip python script that uses these codes.

### DIFF
--- a/src/pcms/pythonapi/bind_uniform_grid_field_layout.cpp
+++ b/src/pcms/pythonapi/bind_uniform_grid_field_layout.cpp
@@ -6,6 +6,49 @@
 
 namespace py = pybind11;
 
+// Mirror pybind11's built-in std::array type caster for Kokkos::Array so that
+// def_readwrite members work with Python sequences (e.g. grid.bot_left = [0,
+// 0]).
+namespace pybind11::detail
+{
+template <typename Type, size_t Size>
+struct type_caster<Kokkos::Array<Type, Size>>
+{
+  using ArrayType = Kokkos::Array<Type, Size>;
+  using value_conv = make_caster<Type>;
+
+  bool load(handle src, bool convert)
+  {
+    if (!isinstance<sequence>(src))
+      return false;
+    auto seq = reinterpret_borrow<sequence>(src);
+    if (seq.size() != Size)
+      return false;
+    size_t i = 0;
+    for (auto it : seq) {
+      value_conv conv;
+      if (!conv.load(it, convert))
+        return false;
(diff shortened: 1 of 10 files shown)


test/test_point_search.cpp
```diff
@@ -139,7 +139,7 @@ TEST_CASE("construct intersection map")
     Kokkos::View<Uniform2DGrid[1]> grid_d("uniform grid");
     auto grid_h = Kokkos::create_mirror_view(grid_d);
     grid_h(0) = Uniform2DGrid{
-      .edge_length{1, 1}, .bot_left = {0, 0}, .divisions = {10, 10}};
+      .edge_length{{1, 1}}, .bot_left = {{0, 0}}, .divisions = {{10, 10}}};
```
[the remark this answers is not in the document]
Fuad-HH: Fixed.

src/pcms/pythonapi/bind_uniform_grid_field_layout.cpp
Fuad-HH, line 1: I had to create this type caster for `Kokkos::Array` support in Python. I am not sure if this is the best way. Added this with AI.

Fuad-HH: Moved it to `numpy_array_transform.h`.

src/pcms/pythonapi/numpy_array_transform.h
```diff
@@ -7,6 +7,48 @@
...
+    for (auto& v : src)
+      l[i++] = make_caster<Type>::cast(v, policy, parent);
+    return l.release();
+  }
+
+  PYBIND11_TYPE_CASTER(ArrayType, const_name("List[") +
+                                    make_caster<Type>::name + const_name("]"));
+};
+} // namespace pybind11::detail
+
```
Fuad-HH, line 51: I had to create this type caster for Kokkos::Array support in Python. I am not sure if this is the best way. Added this with AI.

jacobmerson (approved):

jacobmerson:
Thanks for the fixes. This looks ready to merge once the tests finish.

PR #350, opened 2026-08-05
Conservative transfer 3D

jacobmerson:


abhiyanpaudel:
It looks good to me. I have updated it with an additional 3D test case.

src/pcms/transfer/conservative_projection_solver.cpp
```diff
@@ -56,6 +56,8 @@ GalerkinProjectionSolver::GalerkinProjectionSolver(
   CHKERRABORT(PETSC_COMM_WORLD, ierr);
   ierr = KSPSetFromOptions(ksp_);
   CHKERRABORT(PETSC_COMM_WORLD, ierr);
+  ierr = KSPSetTolerances(ksp_, 1e-14, 1e-14, PETSC_DEFAULT, PETSC_DEFAULT);
```
jacobmerson, line 59: @abhiyan123 why are you hardcoding tolerances instead of using the defaults / passing options?

jacobmerson:
Something strange happened with the merge/rebase @abhiyan123 . Let me see if I can resolve that so we don't have so many unrelated changes...

jacobmerson:
This has some work to deal with the merge with develop due to the meshfields 1.0 transition.

PR #352, opened 2026-08-06
improve code reuse in tests

jacobmerson:
Factor the repeated mesh construction, function-space creation, field sampling and evaluation boilerplate out of the individual test translation units and into field_test_utils.h:

  - BuildUnitSquare / MakeP1Space / MakeP0Space for the unit-square simplex meshes used across the transfer and evaluation tests
  - IntegrateP0Field / IntegrateP1Field for conservation checks
  - EvaluateAndAssemble for the sample-then-assemble pattern shared by the linear form integrator tests
  - StandardOutsideCoords2D for out-of-bounds policy tests

SetField now dispatches on the arity of the supplied callable so the same helper covers 2D and 3D layouts, and it samples the layout's DOF coordinates directly instead of copying them into a fixed-rank view. The evaluation helpers use rank-2 output views through MakeRank2View rather than hand-built Rank2Views over rank-1 storage.

### DIFF
--- a/test/field_test_utils.h
+++ b/test/field_test_utils.h
@@ -3,6 +3,7 @@
 
 #include <catch2/catch_approx.hpp>
 #include <Kokkos_Core.hpp>
+#include "pcms/configuration.h"
 #include "pcms/field/field.h"
 #include "pcms/field/field_data.h"
 #include "pcms/field/field_evaluator_factory.h"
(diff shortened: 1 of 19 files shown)


jacobmerson:
@Sichao25 pretty sure this is mergable, it's just moving shared code out of individual tests and into the helper utils.

